PR #2, opened 2016-08-23
s1965 - changed build from ant to gradle

brianmajor:


opencadc-admin:
Please remove cruft like Dependencies.txt and check javadoc generation

PR #4, opened 2016-08-29
queryFor.... changed

bertocco:


brianmajor:
Sara,

Thank you for the changes, but we have decided to stick with an older version of the spring-jdbc libraries because the changes spread too far.  I had made these very same changes myself which I had to discard.  So, if you get the latest UWS and VOS code the compile should work because we changed back to the old version of spring-jdbc.

Brian

Issue #11, opened 2016-10-18
VOSpaceClient.deleteNode does not allow cookie authentication


at88mph:
The setRequestCookie(HttpConnection) call is missing because it is not consistent with the other Http[PutPostGet] classes.

PR #12, opened 2016-10-18
Issue #11: Authenticated deleteNode() call.

at88mph:
Brought more inline with other HttpXXX classes.  Added call for setRequestSSOCookie().

### DIFF
--- a/cadc-vos/build.gradle
+++ b/cadc-vos/build.gradle
@@ -25,7 +25,7 @@ dependencies {
     compile 'xerces:xercesImpl:2.+'
     compile 'org.json:json:20160212'
 
-    compile 'org.opencadc:cadc-util:[1.0.6,)'
+    compile 'org.opencadc:cadc-util:[1.0.7,)'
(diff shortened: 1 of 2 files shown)


pdowler:
How much work to move the anonymous httpDelete class to cadc-util (net pkg) as a real HttpDelete class? That way work to support new credential types and deal with retries could be added later and all apps would benefit.

at88mph:
Did you look at the pull request?  I’m not sure if those changes pretty much cover it or not; meaning we could transport those changes into an HttpDelete class since my changes make use of overriding specific methods of HttpTransfer.


Dustin N. Jenkins 
Tel/Tél: 250.217.4970 | facsimile/télécopieur: 250.363.0045 | Dustin.Jenkins@nrc-cnrc.gc.ca 
National Research Council Canada | 5071 West Saanich Rd, Victoria BC. V9E 2E7 
Conseil national de recherches Canada | 5071, ch. West Saanich, Victoria (C.-B) V9E 2E7 
Government of Canada | Gouvernement du Canada

PR #20, opened 2017-02-15
bug fix for picking AuthMethod copied from GMSClient

pdowler:


cadc-vos/src/main/java/ca/nrc/cadc/vos/client/VOSpaceClient.java
```diff
@@ -718,22 +720,73 @@ public void write(OutputStream out) throws IOException
...
+                return AuthMethod.COOKIE;
             }
+            
+            // AuthMethod.PASSWORD not supported
+            // AuthMethod.TOKEN not supported
+            throw new AccessControlException("No valid public credentials.");
+        }
+        else
+        {
+            throw new AccessControlException("Anonymous access not supported.");
```
brianmajor: Anonymous access has to be allowed in VOSpace...

pdowler: OK, easy to fix at bottom of getAuthMethod... what about if there is a cookie but wrong domain? drop to anon? That is something that could happen in a UI application... which I guess should try.

brianmajor: I would say that it should fail, otherwise we'll be hiding a domain config problem?

pdowler: reverted the domain mismatch to fail

Issue #10, opened 2016-10-07
CRED_SERVICE_ID should be configurable


bertocco:
The CRED_SERVICE_ID in cadc-vos-server/src/main/java/ca/nrc/cadc/vos/server/auth/VOSpaceAuthorizer should be configurable

Proposed patch 

Move CRED_SERVICE_ID in constructor and use localAuthority.getServiceURI:

public class VOSpaceAuthorizer implements Authorizer
{
    protected static final Logger LOG = Logger.getLogger(VOSpaceAuthorizer.class);

```
// TODO: make these configurable or find from the VOSpace capabilities
//private static final String CRED_SERVICE_ID = "ivo://cadc.nrc.ca/cred";
private static String CRED_SERVICE_ID;


 public VOSpaceAuthorizer()
{
    this(false);

    LocalAuthority localAuthority = new LocalAuthority();
    URI serviceURI = localAuthority.getServiceURI(Standards.CRED_DELEGATE_10.toString());
    CRED_SERVICE_ID = serviceURI.toString();

}

public VOSpaceAuthorizer(boolean allowPartialPaths)
{
    initState();
    this.allowPartialPaths = allowPartialPaths;

    LocalAuthority localAuthority = new LocalAuthority();
    URI serviceURI = localAuthority.getServiceURI(Standards.CRED_DELEGATE_10.toString());
    CRED_SERVICE_ID = serviceURI.toString();
}
```

brianmajor:
The VOSpaceAuthorizer should be changed to use the checkCredentials() method from CredUtil.java (in project cadc-cdp-server).  If the call to checkCredentials() returns false, hasMembership() should also return false.  If it returns true, the membership check should continue.

brianmajor:
Fixed.

PR #21, opened 2017-02-16
Fix for issue #10

yeunga:
Fix for issue #10

### DIFF
--- a/cadc-vos-server/build.gradle
+++ b/cadc-vos-server/build.gradle
@@ -14,7 +14,7 @@ sourceCompatibility = 1.7
 
 group = 'org.opencadc'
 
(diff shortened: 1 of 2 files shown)


brianmajor:
Hi Alinga.  The code looks good except for a few places where the indenting is off (an extra 2 spaces in the if checkCredentials() block.

Also, could you run the vospace integrations tests from (service.git) with this code?  You'll probably want to run them with the old code first to make sure they pass.

yeunga:
Hi Brian - Ran the vospace integration tests. All tests passed except for testCopyFromVOSpace() which sometimes fails, even on the released production codes.

PR #22, opened 2017-03-02
Don't fail getNodeAction if the forward-checking of child permissions check fails

brianmajor:
When calling a remote GMS service, the user's proxy cert may not have a recognized certificate authority.  This should not cause the entire get node action to fail.


brianmajor:
Think you are right.. I've added the more general catch to that loop as well.

PR #24, opened 2017-03-17
Change cadc-vos ClientTransfer to use HttpPost.

hjeeves:


cadc-vos/src/main/java/ca/nrc/cadc/vos/client/ClientTransfer.java
```diff
@@ -90,26 +88,18 @@
 import javax.net.ssl.SSLSocketFactory;
 import javax.security.auth.Subject;
 
+import ca.nrc.cadc.net.*;
```
brianmajor: The convention in the opencadc code is to list each of the imports instead of using wildcard imports.

cadc-vos/src/main/java/ca/nrc/cadc/vos/client/ClientTransfer.java
```diff
@@ -143,10 +135,10 @@ private ClientTransfer() { }
     /**
      * @param jobURL UWS job URL for the transfer job
      * @param transfer a negotiated transfer
-     * @param monitor monitor the job until complete (true) or just start
+     * @param transListener monitor the job until complete (true) or just start
      *  it and return (false)
      */
-    ClientTransfer(URL jobURL, Transfer transfer, boolean schemaValidation)
+    ClientTransfer(URL jobURL, Transfer transfer, boolean transListener)
     {
```
brianmajor: The schemaValidation parameter here has been lost.... seems like it was replaced with transListener which is not used.

cadc-vos/src/main/java/ca/nrc/cadc/vos/client/ClientTransfer.java
```diff
@@ -545,41 +537,33 @@ private void doServerTransfer()
...
-            int responseCode = connection.getResponseCode();
-            String responseMessage = connection.getResponseMessage();
-            String errorBody = NetUtil.getErrorBody(connection);
-            if (StringUtil.hasText(errorBody))
-                responseMessage += ": " + errorBody;
-            
-            switch (responseCode)
+
+            HttpPost transfer = new HttpPost(phaseURL, parameters.toString(),
+                    "application/xml", false);
```
brianmajor: The constructor public HttpPost(URL url, Map<String, Object> map, boolean followRedirects) should be used here instead:
  - the parameters should go in the map
  - since there isn't an XML document being posted there's no need to set the content-type.

cadc-vos/src/main/java/ca/nrc/cadc/vos/client/ClientTransfer.java
```diff
@@ -545,41 +537,33 @@ private void doServerTransfer()
...
+            transfer.setSSLSocketFactory(getSSLSocketFactory());
```
brianmajor: Can't we use the existing initHTTPS() method for this work instead?

PR #37, opened 2017-12-10
removed the forced posixGroup

pdowler:
Using the posix group with rwx to give cavern service full permissions on files also does one of:

- gives all users full permissions to those files (users belong to standard group)
- users can create files via mount that cavern service cannot manipulate (users do not belong to standard group so cannot chgrp)

Basically, the concept is not workable; cavern will need full permissions to the content tree via some other mechanism like ACLs

### DIFF
--- a/cavern/src/main/java/org/opencadc/cavern/FileSystemNodePersistence.java
+++ b/cavern/src/main/java/org/opencadc/cavern/FileSystemNodePersistence.java
@@ -106,7 +106,6 @@ public class FileSystemNodePersistence implements NodePersistence {
 
     private PosixIdentityManager identityManager;
     private Path root;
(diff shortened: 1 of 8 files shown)


pdowler:
Brian - when I ran all the tests it looked like the CavernURLGeneratorTest test was hanging but I did not investigate. Not sure why it has all that RegistryClient property setup if it is a unit test.

brianmajor:
Pat, my mistake--I'd meant to remove the Registry dependency but forgot.  That's fixed now in a separate pull request.

brianmajor:
Regarding the posix group for the service:  yes, I think you're right.  Having that group complicates matters quite a bit anyhow.

PR #39, opened 2018-01-17
cavern improvements for group permissions and file checksums

pdowler:


cavern/src/main/java/org/opencadc/cavern/FileSystemNodePersistence.java
```diff
@@ -233,8 +246,40 @@ public Node put(Node node) throws NodeNotSupportedException, TransientException
...
+                        cur.setMarkedForDeletion(true);
+                    } else { // update
+                        cur.setValue(prop.getPropertyValue());
+                    }
+                } else { // add
+                    if (!prop.isMarkedForDeletion()) {
+                        node.getProperties().add(prop);
+                    } // else: ignore delete non-existent prop
+                }
+            }
```
brianmajor: What about properties whose values have changed?

cavern/src/main/java/org/opencadc/cavern/nodes/AclCommandExecutor.java
```diff
@@ -0,0 +1,191 @@
...
+        grabber.captureOutput(cmd);
+        if (grabber.getExitValue() != 0) {
+            throw new IOException("failed to set read-only ACL on " + path + ": "
+                + grabber.getErrorOutput(true));
+        }
+        String out = grabber.getOutput(true);
+        String[] lines = out.split("[\n]");
+        for (String s : lines) {
+            if (!s.startsWith("#")) {
+                String[] tokens = s.split(":");
```
brianmajor: Think this acl output is fairly standard?

pdowler: I sure hope so :-) All linux systems will use the same command line tools for this. I guess some options would make parsing more robust: -c (omit header) and -s (no output if only base entries present). Something to do in in the refactor/cleanup phase.

cavern/src/main/java/org/opencadc/cavern/FileSystemNodePersistence.java
pdowler: changed values are handled in the cur != null branch above a(at line 261). I'll admit that the FileSystemNodePersistence probably needs unit tests now as it started to have some logic and  not just wrapping calls to NodeUtil...  either that or I put that login in the wrong place. That is something we could decide when we review the public API(s) (after we get everything to work)

cavern/src/main/java/org/opencadc/cavern/nodes/AclCommandExecutor.java
pdowler: waiting for an hour-long job to run so I will improve this a little...

brianmajor:
New changes look good!

PR #40, opened 2018-01-26
Set HTTP Headers on File Download

brianmajor:


cavern/src/intTest/java/org/opencadc/cavern/MetadataIntTest.java
```diff
@@ -118,6 +132,62 @@ public static void staticInit() throws Exception
             throw new IllegalStateException("expected system property " + uriProp + " = <base vos URI>, found: " + uri);
     }
 
+    @Test
+    public void testHeaderMetadata() throws Exception {
```
pdowler: Can you rename this test method (and maybe testContentMD5) so it is clear in a report of pass/fail (eg like we did for c3tp) that this tests the file download header and the other tests the node metadata header?

Maybe testFileDownloadContentMD5() and testNodeContentMD5()??

cavern/src/main/java/org/opencadc/cavern/files/CavernURLGenerator.java
```diff
@@ -173,7 +173,7 @@ public CavernURLGenerator(String root) {
             }
 
             List<URL> baseURLs = getBaseURLs(target, protocol.getSecurityMethod(), scheme);
-            if (baseURLs == null) {
+            if (baseURLs == null || baseURLs.size() == 0) {
```
pdowler: my IDE always whines that I should use isEmpty() instead of size()==0 and it seems like a good idea

cavern/src/main/java/org/opencadc/cavern/files/GetAction.java
```diff
@@ -116,8 +120,18 @@ public void doAction() throws Exception {
                 return;
             }
             
-            // TODO: get node so we can put appropriate properties into HTTP headers
+            // set HTTP headers
+            Path rootPath = Paths.get(getRoot());
+            Node node = NodeUtil.get(rootPath, nodeURI);
```
pdowler: Are you sure this works correctly when there are links in the path? I had to change NodeUtil.get to work with PathResolver and that meant returning a LinkNode when I ran into one... line 439 of NodeUtil. basically, I think the break on line 446 means return null - maybe that needs to change to resolve the link to a container and continue? (eg call pathToNode again but without using LinkOption.NOFOLLOW_LINKS (on line 493)... I'm not sure and the NodeUtilTest isn't thorough enough

cavern/src/intTest/java/org/opencadc/cavern/MetadataIntTest.java
brianmajor: Done

cavern/src/main/java/org/opencadc/cavern/files/CavernURLGenerator.java
brianmajor: Done

cavern/src/main/java/org/opencadc/cavern/files/GetAction.java
brianmajor: I'll have to think about this more.  Yes, more tests are probably needed.

PR #43, opened 2018-01-30
Resolve links on transfer to get MD5

brianmajor:


cavern/gradle.properties
```diff
@@ -2,6 +2,7 @@ systemProp.ca.nrc.cadc.util.PropertiesReader.dir = build/resources/test
 systemProp.org.opencadc.cavern.AuthAccessTest.baseURI = vos://canfar.net~cavern/cadcregtest1/vospace-int-test/auth
 systemProp.org.opencadc.cavern.TransferRunnerTest.baseURI = vos://canfar.net~cavern/cadcregtest1/vospace-static-test/transfer
 systemProp.org.opencadc.cavern.MetadataIntTest.baseURI = vos://canfar.net~cavern/cadcregtest1/vospace-int-test/metadata
+systemProp.org.opencadc.cavern.MetadataIntTest.linksBaseURI = vos://canfar.net~cavern/cadcregtest1/vospace-int-test/metadata-links
```
pdowler: This is the same test with a separate base dir for some test methods?

brianmajor:
Yes, the tests create a link in linksBaseURI that points a node in baseURI
and ensure that the node is resolved and that the MD5 returned is that of
the data node.

cavern/src/main/java/org/opencadc/cavern/files/GetAction.java
```diff
@@ -116,7 +121,18 @@ public void doAction() throws Exception {
...
-            // TODO: get node so we can put appropriate properties into HTTP headers
+            // set HTTP headers.  To get node, resolve links but no authorization (null authorizer) 
+            NodePersistence nodePersistence = new FileSystemNodePersistence();
+            PathResolver pathResolver = new PathResolver(nodePersistence, true);
+            Node node = pathResolver.resolveWithReadPermissionCheck(nodeURI, null, true);
+            log.info("BM: req: " + nodeURI + " res: " + node.getUri() + " md5: " + node.getPropertyValue(VOS.PROPERTY_URI_CONTENTMD5));
+            String contentEncoding = node.getPropertyValue(VOS.PROPERTY_URI_CONTENTENCODING);
+            String contentLength = node.getPropertyValue(VOS.PROPERTY_URI_CONTENTLENGTH);
+            String contentMD5 = node.getPropertyValue(VOS.PROPERTY_URI_CONTENTMD5);
+            syncOutput.setHeader("Content-Disposition", "inline; filename=" + node.getName());
```
pdowler: You should add Content-Type to the props here... user-set like encoding but should be supported

cavern/src/main/webapp/capabilities.xml
```diff
@@ -44,6 +44,10 @@
       <accessURL use="base">http://replace.com/cavern/auth/nodes</accessURL>
       <securityMethod standardID="http://www.w3.org/Protocols/HTTP/1.0/spec.html#BasicAA" />
```
pdowler: We should scrap this non-standard BasicAA  entry since no s/w uses it

brianmajor:
Thanks, I'm surprised I didn't add it.

PR #49, opened 2018-06-04
s1990: support sorting on contentLength, lastModified, asc & desc for vosapce

hjeeves:
Server side of it, at least. There are changes in the python libraries that will be coming as well.


hjeeves:
Closing this in favour of the one bmajor has up currently.

PR #52, opened 2018-06-07
Added Credential Injection servlet

brianmajor:


brianmajor:
Configuration is not portable

PR #53, opened 2018-06-28
Split sort tests from the other container tests for more flexibility in cadc-test-vos

andamian:


cadc-vos-server/src/test/java/ca/nrc/cadc/vos/auth/VOSpaceAuthorizerTest.java
```diff
@@ -187,7 +187,7 @@ public void testNodeLocked() throws Exception
     public void testCheckDelegation() throws Exception
     {
         // create keys
-        RsaSignatureGenerator.main(new String[]{getCompleteKeysDirName()});
+        RsaSignatureGenerator.genKeyPair(getCompleteKeysDirName());
```
pdowler: Yes, this is the right change. I'll accept it here and deal with the merge conflict later.

PR #54, opened 2018-07-05
implement sshfs mount negotiation in cavern

pdowler:
plus related library and test changes


cadc-vos-server/src/main/java/ca/nrc/cadc/vos/server/transfers/TransferGenerator.java
```diff
@@ -65,49 +65,45 @@
...
+public interface TransferGenerator {
 
     /**
-     * Request a list of URLs for the given transfer request information.
+     * Request a list of endpoints for the given transfer request information.
      *
      * This method returns a list of URLs to handle the case where the
      * storage system has multiple copies of a file or multiple locations
-     * in which a file can be saved.  Returning only one URL in the list
+     * in which a file can be saved. Returning only one URL in the list
```
brianmajor: URL->endpoint

cadc-vos-server/src/main/java/ca/nrc/cadc/vos/server/transfers/TransferRunner.java
```diff
@@ -182,9 +182,33 @@ private Transfer createTransfer(List<Parameter> params)
...
+                case TOKEN:
+                    proto.setSecurityMethod(Standards.SECURITY_METHOD_TOKEN);
+                    break;
+                default:
+                    break;
+            }
+        }
+        if (proto.getSecurityMethod() != null) {
+            plist.add(proto);
+        }
```
brianmajor: Is this sequence to help 2.0 clients by adding the most likely intended security method?

cadc-vos-server/src/main/java/ca/nrc/cadc/vos/server/transfers/TransferGenerator.java
pdowler: re-wrote comment to be more up to date

cadc-vos-server/src/main/java/ca/nrc/cadc/vos/server/transfers/TransferRunner.java
pdowler: Yes, it explicitly adds a second protocol with securityMethod that matches the current AuthMethod since the client can't specify securityMethod via param but the service might implement strict 2.1 matching (like cavern does).

PR #56, opened 2018-08-29
Added support for downloading to an outputstream

brianmajor:


cavern/src/intTest/java/org/opencadc/cavern/TransferRunnerTest.java
```diff
@@ -208,6 +246,14 @@ public void testTransferNegotiation21() {
             Assert.assertFalse("cookieTLS", plist.contains(cookieTLS));
             Assert.assertFalse("token", plist.contains(token));
             Assert.assertFalse("tokenTLS", plist.contains(tokenTLS));
+            
+            // check the file content
```
pdowler: Not sure it is useful to do the download here since this test is about expected combinations of protocols and security methods (negotiation only)

cavern/src/intTest/java/org/opencadc/cavern/TransferRunnerTest.java
```diff
@@ -254,6 +300,14 @@ public void testPullFromVOSpace() {
...
                 Assert.assertTrue("one or more endpoints for " + reqP.getUri(), (num > 0));
             }
+            
+            // check the file content
+            trans.setInputStreamWrapper(new InputStreamWrapper() {
+                public void read(InputStream inputStream) throws IOException {
+                    byte[] buf = new byte[100];
+                    Assert.assertEquals(3, inputStream.read(buf, 0, 100));
+                }});
+            trans.runTransfer();
```
pdowler: No actual assert that the bytes received are correct, contrary to the comment. To pass it has to run without failing so that is most of the test. I would declare buf as final outside the anon class and add an assert after the run

PR #57, opened 2018-09-06
Story 2325. Fixed a unit test which failed as a result of changes to …

yeunga:
…JsonInputter and JsonOutputter in core/cadc-util.


yeunga:
For changes to JsonInputter/JsonOutputter, please refer to pull request https://github.com/opencadc/core/pull/61

PR #59, opened 2018-12-10
Story 2346. Added RUNID node property.

yeunga:
This node property is used by DOI. When a DOI web service creates a DOI, it adds this property to the DOI container node the parameter 'runId="TEST"' is present.


cadc-vos/src/main/java/ca/nrc/cadc/vos/VOS.java
```diff
@@ -235,6 +235,9 @@ else if (value.equals(properties.getValue()))
      * Standard Node Properties defined by the CADC
      */
 
+    // Pass information to the service, e.g. runid = "TEST" means the service call is a test
```
brianmajor: Could you please change the comment to:
"Property used for identifying a transaction"

PR #61, opened 2019-01-04
Story 2390. Replaced null contentType with 'text/xml'. A null content…

yeunga:
…Type is no longer valid after Pat's changes.


brianmajor:
Same work was done in the previous pull request.

PR #65, opened 2019-02-21
s2422, isolating vospace tests from use of SSLUtil.initSSL in cadc-test-vos

hjeeves:


cadc-test-vos/src/main/java/ca/nrc/cadc/conformance/vos/VOSTestSuite.java
```diff
@@ -75,6 +75,7 @@
 import java.util.UUID;
```
pdowler: Please deprecate this class - it is still usable but one cannot customise te set of tests they expect to work with their service so we end up writing a different suite class in the service(s).

hjeeves: Deprecate, or remove?

PR #72, opened 2019-08-21
s1630 - added support for Node storageID column.

jburke-cadc:


cadc-vos-server/src/main/java/ca/nrc/cadc/vos/server/NodeDAO.java
```diff
@@ -2718,7 +2723,7 @@ void setValues(PreparedStatement ps)
             sb.append(pval);
             sb.append(",");
 
-            log.debug("setValues:" + sb.toString());
+            //log.debug("setValues:" + sb.toString());
```
brianmajor: did you mean to leave this commented out?

cadc-vos/build.gradle
```diff
@@ -15,7 +15,7 @@ sourceCompatibility = 1.7
 
 group = 'org.opencadc'
 
-version = '1.1.8'
+version = '1.1.9'
```
brianmajor: Looks like there were actually no changes in cadc-vos so you could leave the version as it.  I suspect there was a change originally when you had storageID directly in the Node class.

cadc-vos-server/src/main/java/ca/nrc/cadc/vos/server/NodeID.java
```diff
@@ -82,20 +82,35 @@
...
     /**
      * NodeID constructor.
      * 
      * @param id
      * @param owner
+     * @param ownerObject
+     * @param storageID
      */
-    public NodeID(Long id, Subject owner, Object ownerObject)
+    public NodeID(Long id, Subject owner, Object ownerObject, String storageID)
```
brianmajor: Is this constructor (and the one above with one less param) actually used anywhere?

cadc-vos-server/src/main/java/ca/nrc/cadc/vos/server/NodeDAO.java
jburke-cadc: Yes there is the same debug a bit further down which logs the entire preparedstatement.

cadc-vos/build.gradle
jburke-cadc: You're right. Forgot to rollback the build number when I removed the Node changes. Will fix.

cadc-vos-server/src/main/java/ca/nrc/cadc/vos/server/NodeID.java
jburke-cadc: The 3-arg constructor is used in a unit test. The 4-arg constructor isn't used. I updated assuming both were used in code somewhere, but then never found that somewhere.

cadc-vos/build.gradle
jburke-cadc: Fixed.

cadc-vos-server/src/main/java/ca/nrc/cadc/vos/server/NodeID.java
jburke-cadc: Removed the 4-arg constructor.

PR #78, opened 2020-03-06
Ticket 76375. Fixed handling of limit on the number of children nodes.

yeunga:


cavern/src/test/java/org/opencadc/cavern/nodes/NodeUtilTest.java
```diff
@@ -558,6 +558,34 @@ public void testList() {
...
+                int upperLimit = 0;
+                parent = nn.getParent();
+                iter = NodeUtil.list(root, parent, null, upperLimit);
+                num = 0;
+                while (iter.hasNext()) {
+                    Node nf = iter.next();
+                    log.info("[testList] found: " + nf.getUri());
+                    num++;
+                }
+                Assert.assertEquals("num siblings", upperLimit + 1, num);
```
brianmajor: Hmm, why is one node expected here instead of zero?

yeunga: If I understand the code correctly, num is the number of nodes which includes the parent node as well as the children nodes. So num == 1 accounts for the parent node.

brianmajor: NodeUtil.list() simply returns an iterator of the directory, and num starts at 0, so I don't see where the parent is counted...

cavern/src/main/java/org/opencadc/cavern/nodes/NodeUtil.java
```diff
@@ -702,14 +702,15 @@ public FileVisitResult postVisitDirectory(Path t, IOException ioe)
         try (DirectoryStream<Path> stream = Files.newDirectoryStream(np)) {
             for (Path file : stream) {
                 log.debug("[list] visit: " + file);
+                if (limit != null && (limit == 0 || limit == nodes.size())) {
```
brianmajor: Now that this condition is before the addition of nodes, doesn't the check need to be limit + 1?

Anyhow, I would restore the old code and just check for limit == 0 before the directory stream is created, because we don't need to do that if we don't need any children.

yeunga: Updated code to bail out earlier as suggested when limit = 0.

PR #91, opened 2020-12-17
CADC-8954: provide clientTransfer uws errorSummary check, check HttpUpload failures

hjeeves:
Errors from both these routes were not caught before.  Check for errorSummary is provided as a function so the calling program can decide how to handle errors.

### DIFF
--- a/cadc-vos/src/main/java/ca/nrc/cadc/vos/client/VOSClientUtil.java
+++ b/cadc-vos/src/main/java/ca/nrc/cadc/vos/client/VOSClientUtil.java
@@ -73,13 +73,12 @@ import ca.nrc.cadc.net.ResourceAlreadyExistsException;
 import ca.nrc.cadc.util.StringUtil;
 import ca.nrc.cadc.uws.ErrorSummary;
 import ca.nrc.cadc.uws.ExecutionPhase;
-import ca.nrc.cadc.vos.NodeFault;
+import ca.nrc.cadc.vos.VOS;
 import java.io.FileNotFoundException;
 import java.io.IOException;
 import java.io.StringWriter;
 
 import java.security.AccessControlException;
-import javax.naming.AuthenticationException;
 import org.apache.log4j.Logger;
 import org.jdom2.Document;
 import org.jdom2.output.Format;
@@ -181,8 +180,7 @@ public class VOSClientUtil
     }
 
     public static void checkTransferFailure(ClientTransfer clientTransfer)
-        throws ResourceNotFoundException, IllegalArgumentException,
-        AuthenticationException {
+        throws ResourceNotFoundException, IOException {
         // check to see if anything went wrong
         ErrorSummary errorSummary = null;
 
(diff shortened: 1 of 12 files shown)


brianmajor (changes requested):
- Can you combine checkFailure and checkTransferFailure somehow?  I think the fact that the errors come from different sources should be hidden from users of this class.
- Use the error constants in NodeFault.java for checking the type of transfer error.
- You may as well update this whole class with the correct checkstyle.

brianmajor:
I just realized that NodeFault.java is in cadc-vos-server.  Could you look at moving it to cadc-vos?

brianmajor:
You probably also want to add a helper method in NodeFault.java to make the Fault->Exception mapping easier.

hjeeves:
Moving NodeFault.java turned out to be more problematic than it was worth (would have required an underlying library upgrade that caused a clash with code requiring the older version.) So it's left as is, and will refer to the VOS constants added.

cadc-vos/src/main/java/ca/nrc/cadc/vos/client/VOSClientUtil.java
```diff
@@ -167,4 +179,47 @@ public static void checkFailure(Throwable failure)
...
+            if (StringUtil.hasLength(errorMsg)) {
+                if (errorMsg.contains("Invalid Argument")
+                    || VOS.IVOA_FAULT_INVALID_ARG.equals(errorMsg)) {
+                    log.debug("Invalid Argument found: " + errorMsg);
+                    throw new IllegalArgumentException(VOS.IVOA_FAULT_INVALID_ARG);
+                }
+
+                // not found
+                if (VOS.IVOA_FAULT_NODE_NOT_FOUND.equals(errorMsg) ||
+                    errorMsg.contains("not a data node")) {
```
brianmajor, line 207: Do you still need to check for the non-standard string messages, like "Invalid Argument" and "not a data node"? 

I've seen "not a data node" thrown quite often so perhaps you do.  This is part of a transfer negotiation step I believe, and is thrown when someone is trying to download a directory (container node) and not a file (data node).  So, it should probably be mapped to IllegalArgumentException instead.

cadc-vos/src/main/java/ca/nrc/cadc/vos/client/VOSpaceClient.java
```diff
@@ -364,6 +383,11 @@ public Node setNode(Node node)
         {
             throw new IllegalStateException("Node not found", e);
         }
+        catch (ResourceAlreadyExistsException e)
+        {
+            log.debug("failed to set node", e);
+            throw new IllegalStateException("failed to set node", e);
```
brianmajor, line 389: I'm sure there's a good reason why these all end up as IllegalStateExceptions?  I just don't remember...

brianmajor (changes requested):

cadc-vos/src/main/java/ca/nrc/cadc/vos/client/VOSClientUtil.java
hjeeves: I've also seen 'not a data node' thrown in the past few weeks of testing. And items like 'invalidArgument' (no space) - there's only so much testing for non-standard that can be done.

cadc-vos/src/main/java/ca/nrc/cadc/vos/client/VOSpaceClient.java
hjeeves: As discussed in Slack - I don't know why, just see that it's used throughout cadc-vos and cadc-vos-server.

cadc-vos/src/main/java/ca/nrc/cadc/vos/client/VOSClientUtil.java
hjeeves: Changed 'node not found' to 'IllegalArgumentException'

cadc-vos/src/main/java/ca/nrc/cadc/vos/client/VOSpaceClient.java
hjeeves: Believe the decision is to leave this as-is.

cadc-vos/src/main/java/ca/nrc/cadc/vos/client/VOSClientUtil.java
brianmajor: node not found -> ResourceNotFoundException
not a data node -> IllegalArgumentException

brianmajor (changes requested):

brianmajor (approved):

PR #98, opened 2021-05-07
use latest auth libraries

brianmajor:


brianmajor:
I was going to make the VOSURI based on runtime config, but I found it's deeply embedded in the Restlet code in cadc-access-control-server.  So, I suggest that work is done when we remove Restlet, which will have to be soon in order to pickup the token support.  (@pdowler )

cavern/VERSION
```diff
@@ -0,0 +1,4 @@
+## deployable containers have a semantic and build tag
+# semantic version tag: major.minor
+# build version tag: timestamp
+TAGS="0.2.2 $(date -u +"%Y%m%dT%H%M%S")"
```
pdowler, line 4: we eventually found that having separately tagged images didn't help ops at all, so I have been including a dash between the semantic and build, eg 0.2.2-$(...) so there is one TAG with both bits of info

pdowler (approved):
We'll have to make changes to the testing so it isn't all hard coded to arc, but that's for another day. Just one comment about VERSION below you could address now or later.

cavern/VERSION
brianmajor: I wouldn't want to have to update the k8s deployment info when re-iterating on testing a particular version.  The config refers to, :0.2.2 (for example) and uses the policy to 'Always Pull', meaning it checks for new versions of the ImageID for that tag, but doesn't pull if there are no changes.  I haven't been using the timestamped tag but think it's useful for keeping history.

PR #109, opened 2021-11-25
ACL and cavern improvements

brianmajor:
- Refactored cadc-vos-server so that base VOS URI comes from system property.  If not found defaults to vault.
- Refactored cadc-test-vos to use same system property for base URIs
- Removed test suite from cavern and made subclasses for conformance tests
- Added tests for default ACL behaviour
- Added mask node property
- Change VOSpaceAuthorizer to mask group permissions


cadc-vos-server/src/main/java/ca/nrc/cadc/vos/server/LocalServiceURI.java
```diff
@@ -70,44 +70,38 @@
...
-            props.load(propFile.openStream());
-            String uriString = props.getProperty(URI_KEY);
-            this.serviceURI = new URI(uriString);
-        }
-        catch (Exception e)
-        {
-            log.warn("Unable to read local service URI from " + LOCAL_PROP_FILE +
-                    "\n    Reason: " + e.getMessage(), e);
+        try {
+            String uriBase = System.getProperty(BeanUtil.VOS_BASE_URI_PROP);
```
pdowler, line 95: I agree that a special LocalServiceURI.properties is cumbersome... but this self identity should really come from a config file and be a mandatory config item, so this change is going in the wrong direction. 

We'll have to undo this later when we get to putting all vospace services into containers... so I'm not really saying this needs to be changed now, but it is the wrong direction (experience with SI).

**However, there definitely should not be a default value to one service. That's just wrong and encourages incorrect config so that should be fixed.**

**Secondarily, if you are going to do a blanket catch, don't catch Throwable, catch Exception.** Throwable is also the parent of Error (eg JVM issues like out-of-memory) so you don't want to catch those. It's also bad to catch, log, and then return like nothing happened... but maybe that can't be fixed so easily in code we want to delete anyway.

cadc-vos-server/src/main/java/ca/nrc/cadc/vos/server/util/BeanUtil.java
```diff
@@ -99,6 +98,20 @@ public BeanUtil(final String className)
...
+    /**
+     * Get the base of the VOS URI for this running service.
+     * 
+     * @return The base VOS URI.
+     */
+    public static String getVosUriBase() {
+        String uriBase = System.getProperty(VOS_BASE_URI_PROP);
+        if (uriBase == null) {
+            // default to vault
+            return "vos://cadc.nrc.ca!vault";
```
pdowler, line 111: And in two places no less! ouch

What is this BeanUtil anyway? I've never seen it, no javadoc with an author, and it looks like a complicated way to do simple things. And the whole "bean" thing was a bad idea that died years ago :-)

cavern/VERSION
```diff
@@ -1,4 +1,4 @@
 ## deployable containers have a semantic and build tag
 # semantic version tag: major.minor
 # build version tag: timestamp
-TAGS="0.2.6 $(date -u +"%Y%m%dT%H%M%S")"
+TAGS="0.2.7 $(date -u +"%Y%m%dT%H%M%S")"
```
pdowler, line 4: We discovered that version tags are useful if there is onen tag with the semantic version and the timestamp, so the above should have a dash, eg `0.2.7-$(date ...)`

That makes it much easier to track a running container back to the image and have one tagged image per build.

pdowler (changes requested):

cavern/VERSION
brianmajor: That would make testing in keel-dev a bit of a pain.  The "imagePullPolicy" in k8s that I've found most useful and least dangerous is "Always".  It's the only one that compares the image checksum in the remote registry.  Adding the timestamp to the version would mean the image URI would have to be updated on each change.  Also, the image SHA is there to precisely identify the build.

cadc-vos-server/src/main/java/ca/nrc/cadc/vos/server/util/BeanUtil.java
brianmajor: A very old class tied to the Restlet initialization sequence.  It will go when Restlet goes.  But I'll update it to use LocalServiceURI instead for now.

cadc-vos-server/src/main/java/ca/nrc/cadc/vos/server/LocalServiceURI.java
brianmajor: Okay, I was hesitant about making vault the default, so I'll change it to be a required config item.  I though however that in storage-inventory a lot of java config was moving to catalina.properties, thus the change to a system property.  I think we should get this right now though, so I'll change it back to a file if you think that's the best fit.

cavern/VERSION
brianmajor: I'm open to changing this but this is the way it's done in keel at the moment.  There will be a point I suppose when ops takes over the configuration and they will have to reconcile the different approaches.

cadc-vos-server/src/main/java/ca/nrc/cadc/vos/server/LocalServiceURI.java
pdowler: in SI we're using catalina.properties for things that tomcat has to have values for (essentially, things in server.xml or context.xml) plus some generic/legacy stuff like registry override or that prop to trust client cert in a header (which is library config) but not in general... applications have their own config file and some libs still have a config file as well. This is a lib config file, so probably the right usage for containers.

cavern/VERSION
pdowler: OK, it's really up to operational constraints. But you must not be using either of the tags then, because I only proposed tagging once with both pieces of info instead of twice. How do you revert to a previous image when there is a broken image?

pdowler (approved):

cavern/VERSION
brianmajor: Testing/iteration is done on keel-dev using the same tagged version eg :1.3.  When ready to release, the keel-prod config is changed from 1:2 to 1.3 and can be reverted if needed.  I probably haven't explained this very well!

PR #110, opened 2021-12-14
fix gradle.properties typo

brianmajor:


brianmajor:
Actually I see I can do that.

PR #114, opened 2022-01-10
CADC-10637: Transfer.java, XSD support multiple targets

hjeeves:
VOSpaceClient has been changed internally to support multiple targets. Quick transfers will remain as single targets for now. cadc-vos is not updated to support multiple targets yet (will be when vos-server is updated, so there's something to test with.)


cadc-vos/src/main/java/ca/nrc/cadc/vos/Transfer.java
```diff
@@ -91,76 +88,49 @@
...
-     * @param protocols
+     * Ctor for Transfer. Use this for setting up transfers using multiple targets.
+     * Caller must call getProtocols().add(...) and getTargets().add(...)
+     * @param direction - Direction of transfer (to or from VOSpace)
      */
-    public Transfer(URI target, Direction direction, View view, List<Protocol> protocols)
-    {
-        this.target = target;
+    public Transfer(Direction direction) {
         this.direction = direction;
```
brianmajor, line 122: I don't like the idea of a contructor not having all required items.  The object is in an unusable state after this because at least one target is required.  Why not have a constructor that takes a list?

brianmajor (changes requested):

cadc-vos/src/main/java/ca/nrc/cadc/vos/Transfer.java
hjeeves: Based on previous design discussions, we're going with the constructors as they are, and adding code to instantiate the lists for targets and protocols at variable declaration time.

cadc-vos/src/main/java/ca/nrc/cadc/vos/Transfer.java
```diff
@@ -91,76 +88,46 @@
      */
     public int version = VOS.VOSPACE_20;
 
-    private URI target;
+    private List<URI> targets = new ArrayList<URI>();
```
brianmajor, line 91: This variable must be public.  The following pattern should be followed:

Mandatory fields should be private and have setters and getters.  There's only 1 now: direction.

All others (that are expected to be set by clients) should be public and not have getters and setters.

brianmajor (changes requested):

cadc-vos/src/main/java/ca/nrc/cadc/vos/Transfer.java
```diff
@@ -316,4 +279,28 @@ public String toString()
...
+        this.direction = direction;
+    }
+
+    public void setView(View view) {
+        this.view = view;
+    }
+
+    public void setProtocols(List<Protocol> protocols) {
+        this.protocols = protocols;
+    }
```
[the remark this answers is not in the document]
hjeeves: done.

cadc-vos/src/main/java/ca/nrc/cadc/vos/Transfer.java
hjeeves: As discussed - targets and protocols member variables of Transfer.java are now public, setters and getters removed for those variables. All other member variables of Transfer.java are staying as-is.

cadc-vos/src/main/java/ca/nrc/cadc/vos/Transfer.java
```diff
@@ -91,76 +88,46 @@
...
 
     /**
-     * Constructor for copying a node with a specific view.
-     *
-     * @param target
-     * @param direction
-     * @param view
-     * @param protocols
+     * Ctor for Transfer. Use this for setting up transfers using multiple targets.
+     * Caller must call getProtocols().add(...) and getTargets().add(...)
```
brianmajor, line 116: Yes but it's no longer getProtocols() and getTargets(), it's protocols.add() and targets.add()

cadc-vos/src/main/java/ca/nrc/cadc/vos/Transfer.java
```diff
@@ -254,60 +212,49 @@ public void setRemoteIP(String remoteIP) {
...
      */
-    public List<String> getAllEndpoints()
-    {
+    public List<String> getAllEndpoints() {
         List<String> rtn = new ArrayList<String>();
-        if (this.protocols != null)
-        {
-            for (Protocol p : this.protocols)
-            {
+        if (this.protocols != null) {
```
brianmajor, line 217: Can remove this null check (and the one on targets)

cadc-vos/src/main/java/ca/nrc/cadc/vos/TransferReader.java
```diff
@@ -223,70 +208,67 @@ private Transfer parseTransfer(Document document, String targetScheme)
         if (keepBytesStr != null)
             keepBytes = keepBytesStr.equalsIgnoreCase("true");
 
-        Transfer ret = new Transfer(target, direction, view, protocols, keepBytes);
+        Transfer ret = new Transfer(direction);
+        ret.targets = targets;
```
brianmajor, line 212: As mentioned in slack:  this should be `ret.getTargets().addAdd(targets)` 
So, please change Transfer so that `targets` and `protocols` are private and have getters and setters.

cadc-vos/src/main/java/ca/nrc/cadc/vos/TransferReader.java
```diff
@@ -317,4 +295,29 @@ else if (strDirection.equalsIgnoreCase(Direction.BIDIRECTIONAL.getValue()))
...
+
+    private List<URI> parseTargets(Element root, Namespace vosNS) throws URISyntaxException {
+        List<URI> rtn = null;
+        List targs = root.getChildren("target", vosNS);
+
+        if (targs != null && targs.size() > 0) {
+            rtn = new ArrayList<URI>(targs.size());
+            for (Object obj : targs) {
+                Element eTarget = (Element) obj;
+                URI target = new URI(eTarget.getText());
```
brianmajor, line 307: Maybe call validateTarget() here so there is only one loop?

cadc-vos/src/main/java/ca/nrc/cadc/vos/client/VOSpaceClient.java
```diff
@@ -588,11 +588,16 @@ private ClientTransfer createTransferSync(Transfer transfer)
             HttpPost httpPost = null;
             if (transfer.isQuickTransfer())
             {
-                    Map<String, Object> form = new HashMap<String, Object>();
-                    form.put("TARGET", transfer.getTarget());
-                    form.put("DIRECTION", transfer.getDirection().getValue());
-                    form.put("PROTOCOL", transfer.getProtocols().iterator().next().getUri()); // try first protocol?
-                    httpPost = new HttpPost(vospaceURL, form, false);
+                // Assumption: quick transfer will be a single target
+                if (transfer.targets == null || transfer.targets.size() != 1) {
```
brianmajor, line 592: If targets is null here or if there are zero targets the error message would be confusing.  Maybe just check `if (transfer.targets.size() > 1`... it wasn't checking for null before.

cadc-vos/src/main/java/ca/nrc/cadc/vos/client/VOSpaceClient.java
```diff
@@ -629,8 +634,15 @@ private ClientTransfer createTransferSync(Transfer transfer)
...
             	List<Protocol> prots = new ArrayList<Protocol>();
-            	prots.add(new Protocol(transfer.getProtocols().iterator().next().getUri(), redirectURL.toString(), null));
-            	Transfer trf = new Transfer(transfer.getTarget(), transfer.getDirection(), prots);
+            	prots.add(new Protocol(transfer.protocols.iterator().next().getUri(), redirectURL.toString(), null));
+
+            	// Assumption: quick transfer will be a single target
+                if (transfer.targets == null || transfer.targets.size() != 1) {
+                    throw new IllegalArgumentException("Quick transfer only supports single targets.");
+                }
+
```
brianmajor, line 643: This isn't a 'quick transfer' here so this code shouldn't be here?

cadc-vos/src/main/java/ca/nrc/cadc/vos/client/VOSpaceClient.java
```diff
@@ -629,8 +634,15 @@ private ClientTransfer createTransferSync(Transfer transfer)
...
+                Transfer trf = new Transfer(transfer.targets.get(0), transfer.getDirection());
```
brianmajor, line 644: I think this is the point where you need to get all the targets...

brianmajor (changes requested):

cadc-vos/src/main/java/ca/nrc/cadc/vos/Transfer.java
hjeeves: Except protocols can be set to null in the calling code.?

cadc-vos/src/main/java/ca/nrc/cadc/vos/client/VOSpaceClient.java
hjeeves: Both of these 'Quick transfer' IllegalArgumentException throws are in stanzas starting with: 
            if( transfer.isQuickTransfer()) {

cadc-vos/src/main/java/ca/nrc/cadc/vos/Transfer.java
hjeeves: done

cadc-vos/src/main/java/ca/nrc/cadc/vos/Transfer.java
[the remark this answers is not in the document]
hjeeves: still left this out as it's preferred calling code uses transfer.getProtocols().add() or .addAll(). Same for target list

cadc-vos/build.gradle
```diff
@@ -14,7 +14,7 @@ sourceCompatibility = 1.8
 
 group = 'org.opencadc'
 
-version = '1.1.13'
+version = '1.20'
```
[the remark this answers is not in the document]
hjeeves: set to 2.0

cadc-vos/src/main/java/ca/nrc/cadc/vos/TransferReader.java
hjeeves: combined the validate step into parseTargets.

cadc-vos/src/main/java/ca/nrc/cadc/vos/client/VOSpaceClient.java
hjeeves: removed the null check (now that transfer.targets is private again.) -

cadc-vos/src/main/java/ca/nrc/cadc/vos/TransferReader.java
brianmajor: Looks good

brianmajor (approved):

PR #122, opened 2022-04-27
Cadc 11231 Updated mount and unmount commands in int tests to work on both Mac and Linux platforms.

yeunga:


cavern/VERSION
```diff
@@ -1,4 +1,4 @@
 ## deployable containers have a semantic and build tag
 # semantic version tag: major.minor
 # build version tag: timestamp
-TAGS="0.4.0 $(date -u +"%Y%m%dT%H%M%S")"
+TAGS="0.4.1 $(date -u +"%Y%m%dT%H%M%S")"
```
brianmajor, line 4: No version update required--only tests have changed.

brianmajor (approved):

cavern/VERSION
yeunga: Right. Thanks.

PR #125, opened 2022-05-24
Cadc 44 support quota in storage

yeunga:


cavern/Cavern.properties
```diff
@@ -8,7 +8,6 @@ VOS_FILESYSTEM_ROOT = /cavern
 # 
 SSHFS_SERVER_BASE = proto.canfar.net:/
```
brianmajor, line 9: I know this is just an example file, but could you please change this to ws-uv.canfar.net so that it looks up-to-date?

cavern/src/main/java/org/opencadc/cavern/files/PutAction.java
```diff
@@ -183,15 +185,53 @@ public void doAction() throws Exception {
...
-            UserPrincipal owner = NodeUtil.getOwner(getUpLookupSvc(), node);
-            GroupPrincipal group = NodeUtil.getDefaultGroup(getUpLookupSvc(), owner);
-            NodeUtil.setPosixOwnerGroup(rootPath, target, owner, group);
-            
+            restoreOwnNGroup(rootPath, node);
         } catch (AccessControlException | AccessDeniedException e) {
             log.debug(e);
             syncOutput.setCode(403);
+        } catch (IOException e) {
+            if (e.getMessage().contains("quota exceeded")) {
```
brianmajor, line 193: Maybe just match (case insensitive) on 'quota'.  Not sure how standard that message is but that will give you a slightly better chance of matching on other file systems.

cavern/src/main/java/org/opencadc/cavern/files/PutAction.java
```diff
@@ -183,15 +185,53 @@ public void doAction() throws Exception {
...
+                // TODO: Replace the quota limit below with the number of bytes 
+                //       remaining in the quota when we can determine it.
+                // get quota limit from a parent container
```
brianmajor, line 196: This a good comment, but could you move it closer to where the error message is set?

cavern/src/main/java/org/opencadc/cavern/files/PutAction.java
```diff
@@ -183,15 +185,53 @@ public void doAction() throws Exception {
...
+                Path rootPath = Paths.get(getRoot());
+                restoreOwnNGroup(rootPath, node);
+                Node curNode = node.getParent();
+                String limitValue = curNode.getPropertyValue(VOS.PROPERTY_URI_QUOTA);
+                while (limitValue == null) {
+                    curNode = curNode.getParent();
+                    try {
+                        limitValue = curNode.getPropertyValue(VOS.PROPERTY_URI_QUOTA);
+                    } catch (NullPointerException ex) {
+                        // quota limit not defined in a parent container, exit while loop
```
brianmajor, line 206: Hmm, NullPointerException means an error or bug and shouldn't be part of the program control.  What is null?  Could you check for null instead?

cavern/src/main/java/org/opencadc/cavern/files/PutAction.java
```diff
@@ -183,15 +185,53 @@ public void doAction() throws Exception {
...
+                        break;
+                    }
+                }
+
+                // -1 to indicate quota limit not defined
+                long limit = -1;
+                String msg = e.getMessage();
+                if (limitValue == null) {
+                    // VOS.PROPERTY_URI_QUOTA attribute is not set on the node
+                    msg = "VOS.PROPERTY_URI_QUOTA attribute not set, " + e.getMessage();
```
brianmajor, line 216: This addition to the comment will probably confuse users because it would be a bug in our system.  (quota enforced by not set?). It's good that you're dealing with it though.  I would put a log.warn() message in there instead.

cavern/src/main/java/org/opencadc/cavern/nodes/NodeUtil.java
```diff
@@ -141,7 +141,8 @@ public abstract class NodeUtil {
                         VOS.PROPERTY_URI_GROUPREAD,
                         VOS.PROPERTY_URI_GROUPWRITE,
                         VOS.PROPERTY_URI_ISLOCKED,
-                        VOS.PROPERTY_URI_ISPUBLIC
+                        VOS.PROPERTY_URI_ISPUBLIC,
+                        VOS.PROPERTY_URI_QUOTA
```
brianmajor, line 145: Good eye--this is important to add.

cavern/src/intTest/java/org/opencadc/cavern/QuotaTest.java
```diff
@@ -0,0 +1,200 @@
...
+    @Test
+    public void testQuotaLimitExceeded()
+    {
+        // test case 1: write a small file, should not exceed quota
+        String testFileName = "src/test/resources/quotaTestFileSmall.txt";
+        writeFile(testFileName, false);
+        
+        // test case 2: write a large file that exceeds the quota
+        testFileName = "src/test/resources/quotaTestFile.txt";
+        writeFile(testFileName, true);
```
brianmajor, line 130: If this test fails to throw the exception then the file would still be there.  So, the `writeFile` method should probably delete the node at the end.  You could use the VOSpaceClient to do that.

brianmajor (changes requested):

cavern/Cavern.properties
yeunga: Done

cavern/src/main/java/org/opencadc/cavern/files/PutAction.java
yeunga: Done

cavern/src/main/java/org/opencadc/cavern/files/PutAction.java
yeunga: Done.

cavern/src/main/java/org/opencadc/cavern/files/PutAction.java
yeunga: Done

cavern/src/main/java/org/opencadc/cavern/files/PutAction.java
yeunga: Updated code to look for a null node and log a warning.

cavern/src/intTest/java/org/opencadc/cavern/QuotaTest.java
yeunga: Updated quotaTest to delete the test file at the end of the test.

cavern/src/intTest/java/org/opencadc/cavern/QuotaTest.java
```diff
@@ -177,7 +197,7 @@ public Object run() throws Exception {
                 Throwable ex = paex.getCause();
                 if (expectException && (ex != null) && (ex instanceof IOException)) {
                     String msg = ex.getCause().getMessage();
-                    if (msg.contains("quota exceeded")) {
+                    if (msg.contains("quota")) {
```
brianmajor, line 200:
```suggestion
                    if (msg != null && msg.toLowerCase().contains("quota")) {
```

cavern/src/main/java/org/opencadc/cavern/files/PutAction.java
```diff
@@ -190,35 +190,37 @@ public void doAction() throws Exception {
             log.debug(e);
             syncOutput.setCode(403);
         } catch (IOException e) {
-            if (e.getMessage().contains("quota exceeded")) {
-                // TODO: Replace the quota limit below with the number of bytes 
-                //       remaining in the quota when we can determine it.
-                // get quota limit from a parent container
+            if (e.getMessage().contains("quota")) {
```
brianmajor, line 193:
```suggestion
            String msg = e.getMessage();
            if (msg != null && msg.toLowerCase().contains("quota")) {
```

brianmajor (approved):
Looks good, just those 2 minor comments.

cavern/src/intTest/java/org/opencadc/cavern/QuotaTest.java
yeunga: Updated.

cavern/src/main/java/org/opencadc/cavern/files/PutAction.java
yeunga: Updated.

brianmajor (approved):

PR #130, opened 2022-08-22
CADC-11696 Updated PutAction to clean up orphaned file on put failure.

yeunga:
Did not update integration tests. However all existing integration tests ran successfully.

### DIFF
--- a/cavern/src/main/java/org/opencadc/cavern/files/PutAction.java
+++ b/cavern/src/main/java/org/opencadc/cavern/files/PutAction.java
@@ -236,13 +236,8 @@ public class PutAction extends FileAction {
         log.debug("clean up on put failure " + target);
         Files.delete(target);
         // restore empty DataNode: remove props that are no longer applicable
-        Iterator<NodeProperty> i = node.getProperties().iterator();
-        while (i.hasNext()) {
-            NodeProperty np = i.next();
-            if (VOS.PROPERTY_URI_CONTENTMD5.equals(np.getPropertyURI())) {
-                i.remove();
-            }
-        }
+        NodeProperty npToBeRemoved = node.findProperty(VOS.PROPERTY_URI_CONTENTMD5);
+        node.getProperties().remove(npToBeRemoved);
         NodeUtil.create(rootPath, node);
         return;
     }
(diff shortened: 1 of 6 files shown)


cavern/src/main/java/org/opencadc/cavern/files/PutAction.java
```diff
@@ -227,8 +224,27 @@ public void doAction() throws Exception {
...
+                cleanUpOnFailure(target, node, rootPath);
+            }
+        }
+    }
+    
+    private void cleanUpOnFailure(Path target, Node node, Path rootPath) throws IOException {
+        log.debug("clean up on put failure " + target);
+        Files.delete(target);
+        // restore empty DataNode: remove props that are no longer applicable
+        Iterator<NodeProperty> i = node.getProperties().iterator();
```
brianmajor, line 239: The class Node.java has a method getPropertyValue() you can use so that you don't need to iterate through the properties.

cavern/src/main/java/org/opencadc/cavern/files/PutAction.java
```diff
@@ -227,8 +224,27 @@ public void doAction() throws Exception {
...
+        while (i.hasNext()) {
+            NodeProperty np = i.next();
+            if (VOS.PROPERTY_URI_CONTENTMD5.equals(np.getPropertyURI())) {
+                i.remove();
+            }
         }
+        NodeUtil.create(rootPath, node);
```
brianmajor, line 246: I know this was here before, but why do we create the node again?

brianmajor (changes requested):

cavern/src/main/java/org/opencadc/cavern/files/PutAction.java
yeunga: Yeah, I am just refactoring the existing code. Since we won't be running extensive tests on the updated PutAction, I tried not to make significant changes.

yeunga: If PutAction is always invoked as a result of a adding a node, then we can remove the NodeUtil.create(...) statement. If PutAction can be invoked as a result of an node update, then we need to keep the statement.

yeunga: Keeping this statement to support the (to be implemented) 'feature' to allow uploads without having to create the node first.

cavern/src/main/java/org/opencadc/cavern/files/PutAction.java
yeunga: Thanks for pointing that out. Will use 
`public NodeProperty findProperty(String propertyURI) `
which seems more appropriate for the occasion.

brianmajor (approved):

PR #132, opened 2022-10-04
CADC-11861: update GET and PUT intTests for files endpoint

hjeeves:
plus code tidying - remove unused variables, whitespace, etc.

### DIFF
--- a/cavern/src/main/java/org/opencadc/cavern/files/FileAction.java
+++ b/cavern/src/main/java/org/opencadc/cavern/files/FileAction.java
@@ -162,14 +162,25 @@ public abstract class FileAction extends RestAction {
 
         // Initialize the nodeURI value. Check authorization, either token validation or
         // user authentication against node attributes.
-        String path = syncInput.getPath();
-        if (isPreauth == true) {
-            initPreauthTarget(path);
-        } else {
-            initAuthTarget(path);
+        // Exceptions from init functions are put to syncOutput so they can
+        // be reported to the caller.
+        try {
+            String path = syncInput.getPath();
+            if (isPreauth == true) {
+                initPreauthTarget(path);
+            } else {
+                initAuthTarget(path);
+            }
+        } catch (NodeNotFoundException e) {
+            log.debug("node not found: " + e.getMessage());
+            sendError(404, e.getMessage());
+        } catch (LinkingException e) {
+            log.debug("linking exception: " + e.getMessage());
+            sendError(400, e.getMessage());
+        } catch (IllegalArgumentException e) {
+            log.debug("illegal argument exception: " + e.getMessage());
+            sendError(400, e.getMessage());
         }
-
-        log.debug("sync input available: ");
     }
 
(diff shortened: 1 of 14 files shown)


cavern/src/main/java/org/opencadc/cavern/files/FileAction.java
```diff
@@ -162,14 +162,25 @@ public void initAction() throws Exception {
...
+        // Exceptions from init functions are put to syncOutput so they can
+        // be reported to the caller.
+        try {
+            String path = syncInput.getPath();
+            if (isPreauth == true) {
+                initPreauthTarget(path);
+            } else {
+                initAuthTarget(path);
+            }
+        } catch (NodeNotFoundException e) {
```
brianmajor, line 174: It is correct that you are catching these vospace specific exceptions, but instead of doing the error handling manually, you can just rethrow the mapped exception and let RestAction.java handle them.  So there you would throw a new ResourceNotFoundException(e), and in the next block a new IllegalArgumentException(e).  Shouldn't need the last catch for IllegalArgumentException.

brianmajor: If that works you can get rid of sendError()

cavern/src/main/java/org/opencadc/cavern/files/GetAction.java
```diff
@@ -137,13 +143,14 @@ public void doAction() throws Exception {
...
-        } catch (FileNotFoundException | NoSuchFileException e) {
-            log.debug(e);
-            syncOutput.setCode(404);
-        } catch (AccessControlException | AccessDeniedException e) {
-            log.debug(e);
-            syncOutput.setCode(403);
+        }
+        catch (NodeNotFoundException e) {
+            log.debug("404 error with GET: ",  e);
+            sendError(404, e.getLocalizedMessage());
```
brianmajor, line 149: Same here, can just throw ResourceNotFoundException.

Curious about the removal removal of FileNotFound & NoSuchFileException though...

brianmajor (changes requested):
Should try to work towards the removal of the sendError method.

Also, swagger doc files need to be included.

cavern/src/main/java/org/opencadc/cavern/files/FileAction.java
hjeeves: think this is all done now.

cavern/src/main/java/org/opencadc/cavern/files/GetAction.java
hjeeves: added those other exceptions back in as at this point I don't know if they are run time exceptions or not. Added a check for container nodes because Files.copy was throwing and IOException for action on a container, which was coming out as a 500.

brianmajor (approved):

brianmajor:
Looking good Helena--are int tests passing on rc-uv?  If so I'll merge this and we can work towards release.

hjeeves:
> 

Yes, int tests just finished successfully against rc.

PR #133, opened 2022-10-13
CADC-11904: support '!' and '~' in URI authority in preauth tokens

hjeeves:
Test added into TransferRunnerTest to check that URIs with both '~' and '!' in the authority are supported. Needed because the python code uses '!' (and other clients may as well.)  

Note change to VOSURI.java will require cadc-vos to be released as well as cavern code that needs reviewed before release.

### DIFF
--- a/.github/workflows/gradle.yml
+++ b/.github/workflows/gradle.yml
@@ -11,10 +11,10 @@ jobs:
 
     steps:
     - uses: actions/checkout@v2
-    - name: Set up JDK 1.11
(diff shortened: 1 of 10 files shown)


brianmajor:
@hjeeves You should update the VERSION file for this fix.

brianmajor (changes requested):
Just the one comment about upping the container version.

brianmajor (approved):

PR #135, opened 2022-11-01
Forked `vos.server.db` (Cadc 9058)

andamian:


.github/workflows/gradle.yml
```diff
@@ -19,6 +19,9 @@ jobs:
     - name: build and test cadc-vos
       run: cd cadc-vos && ../gradlew --info clean build javadoc install
       
+    - name: build and test cadc-vos-server-db
```
pdowler, line 22: this needs to be after cadc-vos-server in the build sequence since it implements the NodePersistence interface in there

cadc-vos-server-db/build.gradle
```diff
@@ -0,0 +1,34 @@
...
+group = 'org.opencadc'
+
+version = '1.3.4'
+
+description = 'OpenCADC VOSpace server library DB'
+def git_url = 'https://github.com/opencadc/vos'
+
+dependencies {
+    compile 'org.opencadc:cadc-util:[1.6,2.0)'
+    compile 'org.opencadc:cadc-vos:[1.2.1,)'
```
pdowler, line 26: this should have to depend on cadc-vos-server explicitly... curious why it builds without that

cadc-vos-server-db/src/intTest/java/ca/nrc/cadc/vos/server/NodeDAOTest.java
```diff
@@ -67,7 +67,7 @@
 ************************************************************************
 */
 
-package ca.nrc.cadc.vos.server;
+package ca.nrc.cadc.vos.server.db;
```
pdowler, line 70: package OK - file in wrong directory

pdowler (changes requested):
prelim review with some org and build issues

.github/workflows/gradle.yml
pdowler: ah, you moved NodePersistence to the new module.... that's wrong. It needs to stay here so that cavern, for example, can implement it and not include cadc-vos-server-db in the classpath

cadc-vos-server-db/build.gradle
pdowler: so after above fix, cadc-vos-server-db depends on cadc-vos-server and not the other way around;

then cavern will not include the -db code at all

cadc-vos-server-db/src/main/java/ca/nrc/cadc/vos/server/db/NodeDAO.java
```diff
@@ -733,7 +734,7 @@ private void getOwners(Node node, boolean resolve)
             return;
 
         NodeID nid = (NodeID) node.appData;
-        if (nid.owner != null)
+        if (nid.getOwner() != null)
```
pdowler, line 737: would prefer to keep the old style (ref field rather than getter) since fields are public now... mostly for fewer changes but also because better class design doesn't have public fields and redundant getters

same on line 2742

please revert these shanges

cavern/build.gradle
```diff
@@ -66,7 +66,7 @@ dependencies {
     compile 'org.opencadc:cadc-cdp:[1.0,)'
     compile 'org.opencadc:cadc-gms:[1.0.0,)'
     compile 'org.opencadc:cadc-vos:[1.2.3,)'
-    compile 'org.opencadc:cadc-vos-server:[1.3,)'
+    compile 'org.opencadc:cadc-vos-server:[1.3.4,)'
```
pdowler, line 69: would prefer to not have any changes in cavern, even cosmetic ones

pdowler (changes requested):

cavern/build.gradle
pdowler: also thew change in FilesAction should be reverted

andamian: Done

cadc-vos-server-db/src/main/java/ca/nrc/cadc/vos/server/db/NodeDAO.java
andamian: Done

pdowler (approved):

PR #141, opened 2023-01-26
CADC-11459 Node & NodeProperty updates for SI vospace model

jburke-cadc:


cadc-vos/src/main/java/ca/nrc/cadc/vos/LinkNode.java
```diff
@@ -83,53 +84,39 @@ public class LinkNode extends Node {
...
         return target;
     }
-    
+
     /**
      * Sets the target URI of this LinkNode instance.
-     * 
-     * @param target target URI.
+     *
+     * @param target The URI to the node resource.
```
pdowler, line 119: setTarget needs to enforce the not-null constraint

cadc-vos/src/main/java/ca/nrc/cadc/vos/DataNode.java
```diff
@@ -82,54 +82,75 @@
...
-    /**
-     * Data node constructor.
-     * @param uri
-     */
-    public DataNode(VOSURI uri) {
-        super(uri);
-        this.structured = false;
-        this.busy = NodeBusyState.notBusy;
-    }
+    // The URI to the associated artifact.
```
pdowler, line 86: comments on fields not needed

cadc-vos/src/main/java/ca/nrc/cadc/vos/DataNode.java
```diff
@@ -82,54 +82,75 @@
...
+     * @param name The name of the node.
+     * @param storageID The URI of associated artifact.
+     * @param contentChecksum The artifact checksum.
+     * @param contentLastModified The artifact lastModified date.
+     * @param contentLength The artifact contentLength.
+     * @param properties The node's properties.
      */
-    public boolean isStructured() {
-        return this.structured;
+    public DataNode(String name, URI storageID, URI contentChecksum, Date contentLastModified,
```
pdowler, line 127: We do need two ctors, but not one with properties for reasons given in Node. Leave properties out.

Since the Node->Artifact reference is 0..1, there can be a Node without an artifact and hence all the content* fields can be null. The storageID is never null and is immutable. So, you really need a ctor like:
`DataNode(String name, URI storageID)` 
that sets those to null (because final) and the one you have for when there is an artifact, where I'd prefer 
`DataNode(String name, URI contentChecksum, Date contentLastModfied, Long contentLength, URI storageID)` 
since storageID is at the end because it's really internal. And yes, `Long` instead if `long` and a null check.

cadc-vos/src/main/java/ca/nrc/cadc/vos/ContainerNode.java
```diff
@@ -82,86 +83,53 @@
...
     }
 
     /**
+     * ContainerNode constructor.
      *
-     * @param containerNode
-     * @return true if given container heirarchy matches.
-     * @deprecated
+     * @param name The name of the node.
+     * @param properties The node's properties.
```
pdowler, line 108: remove the ctor with properties.

pdowler (changes requested):

cadc-vos/src/main/java/ca/nrc/cadc/vos/Node.java
```diff
@@ -83,229 +84,134 @@
...
      * on the server side in order to maintain support for older clients.
      */
     public int version = VOS.VOSPACE_20;
-    
-    // The node uri
-    protected VOSURI uri;
 
     // The name of the node
-    protected String name;
+    private String name;
```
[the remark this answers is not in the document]
jburke-cadc: Cleaned up the comments.

cadc-vos/src/main/java/ca/nrc/cadc/vos/Node.java
```diff
@@ -83,229 +84,134 @@
...
      */
-    public Node(VOSURI uri, List<NodeProperty> properties) {
-        this.uri = uri;
-        this.properties = properties;
-
-        this.name = uri.getName();
-        log.debug("uri: " + uri + " -> name: " + name);
-        accepts = new ArrayList<URI>();
-        provides = new ArrayList<URI>();
+    protected Node(String name, Set<NodeProperty> properties) {
```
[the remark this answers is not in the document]
jburke-cadc: Removed.

cadc-vos/src/main/java/ca/nrc/cadc/vos/NodeProperty.java
```diff
@@ -67,104 +67,98 @@
...
  * A VOSpace property representing metadata for a node.
  * 
  * @author majorb
  *
  */
-public class NodeProperty {
-    
-    // Maximum number of groups allowed
-    static final int MAX_GROUPS = 4;
+public class NodeProperty implements Comparable<NodeProperty> {
```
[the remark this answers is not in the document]
jburke-cadc: Added.

cadc-vos/src/main/java/ca/nrc/cadc/vos/NodeProperty.java
```diff
@@ -67,104 +67,98 @@
...
      * Property constructor.
      * 
-     * @param uri The property identifier.
+     * @param key The property identifier.
      * @param value The property value.
      */
-    public NodeProperty(String uri, String value) {
-        this.propertyURI = uri;
-        this.propertyValue = value;
+    public NodeProperty(URI key, String value) {
```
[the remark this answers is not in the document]
jburke-cadc: Updated setValue() to set markedForDeletion to false and added a TODO.

cadc-vos/src/main/java/ca/nrc/cadc/vos/NodeProperty.java
```diff
@@ -67,104 +67,98 @@
...
  * @author majorb
  *
  */
-public class NodeProperty {
-    
-    // Maximum number of groups allowed
-    static final int MAX_GROUPS = 4;
+public class NodeProperty implements Comparable<NodeProperty> {
+
+    // true if the property cannot be modified.
```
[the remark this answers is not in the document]
jburke-cadc: Cleaned up the comments.

cadc-vos/src/main/java/ca/nrc/cadc/vos/LinkNode.java
jburke-cadc: Added a null check.

cadc-vos/src/main/java/ca/nrc/cadc/vos/DataNode.java
jburke-cadc: Cleaned up.

cadc-vos/src/main/java/ca/nrc/cadc/vos/ContainerNode.java
jburke-cadc: Removed that constructor.

pdowler (changes requested):
DataNode and the Structured and Unstructured subclasses also need a ctor with just `String name` argument to be usable client-side and in the NodeReader... I have a small desire for storageID to be immutable (because we don't intend to change it in vault context), but maybe that's not something that's truly generic: it can be null in several contexts.

So, remove the storageID arg from all these ctors and make the field public and not final in DataNode.

cadc-vos/src/main/java/ca/nrc/cadc/vos/ContainerNode.java
```diff
@@ -82,86 +81,43 @@
...
-        boolean structured = i.hasNext();
-        while (i.hasNext() && structured) {
-            if (!i.next().isStructured()) {
-                structured = false;
-            }
-        }
-        return structured;
-    }
+    // The list of child nodes.
+    public final transient List<Node> nodes = new ArrayList<>();
```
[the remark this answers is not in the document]
jburke-cadc: Updated to be private with a public get method.

pdowler:
The model part looks good w.r.t. the UML. With your next commits, can you re-enable CI and have it build the modules that still build (presumably just cadc-vos)? If it's easier I can remove the disable draft stuff in master ... wasn't a good idea in practice.

PR #147, opened 2023-03-27
CADC-12253: switch cavern to use uws postgres job persistence

hjeeves:
rather than in-memory.  README updated to include catalina.properties changes. 

Small changes made in order for base int tests to run:
- location of cadc-tomcat pull updated
- use of Paths changed for use of Files class in AuthPutActionTest because this makes the int tests use resources from the correct build directory (instead of from the src directory) - test had been failing out of the box.

### DIFF
--- a/cavern/src/main/java/org/opencadc/cavern/JobManager.java
+++ b/cavern/src/main/java/org/opencadc/cavern/JobManager.java
@@ -74,7 +74,6 @@ import ca.nrc.cadc.uws.server.SimpleJobManager;
 import ca.nrc.cadc.uws.server.impl.PostgresJobPersistence;
 import org.apache.log4j.Logger;
 
-//public abstract class JobManager extends SimpleJobManager
 public abstract class JobManager extends SimpleJobManager
 {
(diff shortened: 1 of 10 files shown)


cavern/src/main/java/org/opencadc/cavern/JobManager.java
```diff
@@ -67,11 +67,14 @@
...
 package org.opencadc.cavern;
 
+import ca.nrc.cadc.auth.AuthenticationUtil;
 import ca.nrc.cadc.uws.server.JobPersistence;
 import ca.nrc.cadc.uws.server.SimpleJobManager;
 
+import ca.nrc.cadc.uws.server.impl.PostgresJobPersistence;
 import org.apache.log4j.Logger;
 
+//public abstract class JobManager extends SimpleJobManager
```
at88mph, line 77: Delete this unless it's needed for reference.

at88mph (changes requested):

brianmajor (approved):

PR #148, opened 2023-03-28
add entity and re-org packages

pdowler:


pdowler:
One note: the `parent ContainerNode`  in Node is still TBD, but it seems that this might end up still being the simplest way to implement the NodeDAO.... so take that as "probably".

PR #151, opened 2023-04-12
Initial implementation of cadc-vospace-server 2.0 (CADC 12360)

andamian:
New implementation of `cadc-vos-service` library that uses a new `NodePersistence` API.


cadc-vos/src/main/java/org/opencadc/vospace/Node.java
```diff
@@ -211,4 +212,33 @@ public String getPropertyValue(URI uri) {
         return null;
     }
 
+    /**
```
pdowler, line 215: Don't really want convenience methods in the "data model" classes unless the logic has to be internal... these are either wrong or return different results depending on context: please remove from Node.

pdowler (changes requested):
Since we decided to use a feature branch and already have cadc-vos version=2.0 I thought it was obvious that cadc-vos-server would also be version-2.0 and not a separate module.

Please move the new actions and related code into `cadc-vos-server` and make it version=2.0

Also, `vault` is a module in storage-inventory so does not belong here at; the base is in master but there is now a `vos2` branch in storage-inventory  with the current code and that's where integration will happen.

Please remove vault from the PR.

Preferrably: we don't want the history to have commits with junk that was never in master, so that might mean making a new clone and manually copying the files into the right place in the vos2 branch.

pdowler (approved):
merge and fix

PR #152, opened 2023-04-18
CADC-11903 Switched to a different Standard ID that did not use preauth.

yeunga:
Discovered that sssd and acl were not available while running intTest on cavern, Added command to Dockerfile to install sssd and acl packages. If these packages are better installed in cadc-tomcat, then the installation commands will need to be moved to cadc-tomcat.

### DIFF
--- a/cavern/Dockerfile
+++ b/cavern/Dockerfile
@@ -1,7 +1,7 @@
 FROM images.opencadc.org/library/cadc-tomcat:1
 
 ## cavern specific
-RUN dnf install -y sssd-client acl 
+RUN dnf install -y sssd-client acl && dnf clean all
 
 # cavern needs:
 # - sssd-client 
--- a/cavern/Dockerfile
+++ b/cavern/Dockerfile
@@ -1,6 +1,7 @@
 FROM images.opencadc.org/library/cadc-tomcat:1
 
 ## cavern specific
+RUN dnf install -y sssd-client acl 
 
 # cavern needs:
 # - sssd-client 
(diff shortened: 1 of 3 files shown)


brianmajor (changes requested):
The work in CADC-11903 is to use the /files stdID when the user has requested (via the securityMethod field in the transfer) to use a specific security method such as cert or token.  Otherwise the generated URLs should continue to build preauthorized urls (to /preauth).

yeunga:
Associated PR to skaha to deploy to correct version of cavern is [here](https://github.com/opencadc/science-platform/pull/453).

brianmajor:
Hi @yeunga - Because you have loop in there that looks for `"preauth"`, it means we can't change that URL in one place (capabilities.xml) without having to change it in the code as well.  It is a bit of an awkward situation though.  You should only have to look for the stdID for preauth URLS (`vos://cadc.nrc.ca~vospace/CADC/std/archive#file-1.0`).  However, to me, this stdID doesn't communicate that a token needs to be added to it.

Perhaps @pdowler has some thoughts on this.  This issue IMO should be considered low-priority so we chip away at this in the background.

pdowler:
The implementation was complex here because of the arbitrary path length after `/files/` making it hard to detect and extract the preauth token (unlike in minoc, where it is much easier). I have a use case for a securityMethod denoting "token embedded in URL", but not a separate standardID... I don't think that helps here.

I assume the primary difficulty is in CavernURLGenerator generating an absolute URL to /arc/files and it currently reads the capabilities (because a UWS Job only has requestPath).... that's extra bad if the get capabilities uses a registry self-lookup. A good solution doesn't involve self-lookup, so doesn't involve standardID.

thinking about it...

pdowler:
I think we will fix this in the `portable-cavern` branch so I'm closing this.

PR #157, opened 2023-05-10
Disable zip

at88mph:
Added a 503 Servlet for PKG requests to Cavern as they do not check with the underlying permissions.


pdowler:
will review next week, but until then please add some explanation here on the issue and why this is the solution.

at88mph:
I believe it's linked to https://cadc-ccda.atlassian.net/browse/CSD-3452 (ARC permissions)

One can request a ZIP file while not logged in to private files.  The other Download Manager (e.g. URL List) options return a 401 appropriately, but the ZIP download returns a ZIP with the requested files.  I won't put an example here for security reasons.

pdowler:
This is not the right solution for this. The correct solution is to do authorization checks in the loop over files in the FileVisitor... should this PR just be closed?

pdowler:
oh, this is a hack to disable only, not a check and legit response... hmmm

at88mph:
The UI has been disabled for ZIP downloads, but the problem is somebody savvy enough to use CLI tools could conceivably issue a ZIP request for proprietary Cavern URIs.  Once that's fixed, then this PR is irrelevant.

The proper fix is to correct the checking in Cavern.  This can be closed as it's disabled _enough_ for now.

pdowler:
thnx

PR #161, opened 2023-06-13
Support multiple Read and Read/Write Groups

at88mph:


pdowler (commented):
The change in return value for AclCommandExecutor to List looks good.

Since this was wring before (returning a single GroupPrincipal) I'm also questioning the thinking (my own) behind the set methods taking a single GroupPrincipal. It looks to me like `NodeUtil.setNodeProperties` can only clear all ACLs if the property is present (lines 308 and 334) but empty and otherwise only does a set of all groups present - so it won't change an ACL from one group to another.

The VOSpace spec says the group-read and group-write props and multi-valued (space-separated list of strings that are group URIs), so if a node has 
```
ivo://ivoa.net/vospace/core#groupread = ivo://example.net/gms?FOO 
```
and the input node for updating props has
```
ivo://ivoa.net/vospace/core#groupread = ivo://example.net/gms?BAR 
```
this will add an ACL for BAR but not remove the one for FOO. That's the other part of the buggy group permissions. This could be fixed by 
1. always clearing before setting?
2. get ACLs first, figure out ACL(s) to add/remove and execute the minimal commands (more AclCommandExecutor API)
Solution 1 could be easily done in NodeUtil... solution 2 seems more complex and the read-check-write pattern makes it subject to race conditions (other simultaneous updates) - well, more subject to because clear+set can also be hit with that. 

I would probably go with solution 1 (clear-and-set). Thoughts?

pdowler:
The only issue is differentiating between the caller setting a property to null vs not setting it (leaving it as-is), but that's already handled (I think) by the check that the property is present. If the property is present, the call should be setting the whole list of values, not just adding new values to the list.

That does mean that a client (eg storage UI) has to get the current value and modify it (since it is a list) rather than just send in a single new value... and be able to present multiple values to the user so than can add/remove individual values from the list. If that means development work is needed there then "fixing" the bugs on the server side could be dangerous... **well, the storage UI already behaves oddly with vault: it displays multiple group names (/pdowler/junk folder), but won't let me add a 3rd and it won't let me cghange it to a different 2, but it does let me "save" the existing two (so doesn't validate that non-change?).** 

Looking at NodeProperty (master branch), that class is really quite poorly designed. In the `vos2` branch we have removed some of the convenience methods that encourage poor usage, but we haven't integrated enough to know what approach is best (so I can't advise on how to do this so that porting cadc-vos-2.0 will be easy)... but there is a risk of wasting effort trying to do it right in the short term of the PR.

pdowler:
Yes, I agree that it's probably best to make sure the clients can handle multiple groups correctly before changing service behaviour. Can you make this a draft PR to indicate the "pause" status?

at88mph:
Tests fixed (GroupURI _requires_ query parameter for group name now).
Reworked to use the `--set` feature to set ACLs.  Using the `-b` switch wipes out default groups as well, so it is too destructive.

pdowler:
some fixes needed for spec compliance and a bug that NodeUtilTest uncovered;  I have made a PR against your fork/branch.

at88mph:
Integration tests are fixed now and this should be ready for review.

PR #162, opened 2023-08-04
cadc-vos: revert NodeWriter.write(Element) to protected

pdowler:


cadc-vos/src/main/java/org/opencadc/vospace/ContainerNode.java
```diff
@@ -86,7 +87,10 @@ public class ContainerNode extends Node {
     public boolean inheritPermissions = false;
 
     // The list of child nodes.
-    private final transient List<Node> nodes = new ArrayList<>();
+    private final transient List<Node> childNodes = new ArrayList<>();
+    
+    // server side child node iterator
+    public transient ResourceIterator<Node> childIterator;
```
andamian, line 93: By default, this could be the iterator of `childNodes` and overridden on the server side for streaming. `NodeWriter` will use the iterator all the time.

andamian (approved):

PR #166, opened 2023-08-15
Optimize GMS calls (CADC-12708)

andamian:
It contains both optimizations to GMS calls (CADC-12708) and the details attribute in the Node writer/reader(CADC-12725)


pdowler (changes requested):
One of the main features of the GMS-1.0 API is that the search takes one ore more group names and returns the subset the caller belongs to. The IvoaGroupClient method
```
public Set<GroupURI> getMemberships(Set<GroupURI> uris)
```
checks for membership in the set of groups with the minimum number of calls (eg groups them by resourceID and makes one call to each gms service).

So, in VOSpaceAUthorizer there should be no need for a loop over a set of groups, eg VOSpaceAuthorizer line 260. Yes, there is a type-mismatch between Node and IvoaGroupClient (URI vs GroupURI)... 

I think we can change Node to explicitly use GroupURI and then it becomes trivial to just use the set from the node in the call. That would be better than making a new Set every time `hasMembership(Set<URI> groups, Subject subject)` is called and much better than the loop that can cause multiple calls to the gms service instead of one.

The downside of that change (in Node) is that NodePersistence impl code will have to map local group names to absolute GroupURI immediately when constructing the Node, rather than be able to put something simple like `name:CADC` there and make it a GroupURI later, right before output. I don't know what cavern does now, but this is not an issue for vault because vault persists the complete uri anyway.

andamian:
I could not get through with changing the types of the groups sets in the `Node` class because the `GroupURI` class is not a primitive that the `Entity`'s checksum recognizes. We can discuss alternatives.

cadc-vos/src/main/java/org/opencadc/vospace/Node.java
```diff
@@ -210,17 +209,29 @@ public void setName(String name) {
...
         return readWriteGroup;
     }
 
     public Set<NodeProperty> getProperties() {
         return properties;
     }
+
+    @Override
+    protected boolean isDataModelClass(Class c) {
+        if (super.isDataModelClass(c)) {
```
pdowler, line 226: I think I would re-org this to check GroupURI first and if check is false, just return super.isDataModelClass(...) at the end.

for the check, hard-coding the class name will break from refactoring or a typo that the compiler won't catch.

Like this: 
```
if (GroupURI.class.equals(c)) {
  return true;
}

return super.isDataModelClass(c);
```
would be better.

pdowler (changes requested):

cadc-vos/src/main/java/org/opencadc/vospace/Node.java
pdowler: also maybe a comment like
```
// imported data model components
if (GroupURI.class ...
```

andamian: True on both counts. Changed made.

pdowler (approved):

PR #167, opened 2023-08-17
Support for `detail` param in `NodeWriter` (CADC-12725)

andamian:


pdowler:
same content as PR #166

PR #168, opened 2023-08-28
Cadc 12735 Moved linux-specific code to the new cadc-util-fs library

yeunga:


cadc-util-fs/build.gradle
```diff
@@ -0,0 +1,32 @@
...
+    id 'maven'
+    id 'maven-publish'
+    id 'checkstyle'
+}
+
+repositories {
+    mavenLocal()
+    mavenCentral()
+    maven {
+        url "https://maven.restlet.talend.com"
```
pdowler, line 12: remove this extra repo (only needed by restlet)

cadc-util-fs/build.gradle
```diff
@@ -0,0 +1,32 @@
...
+group = 'org.opencadc'
+
+version = '1.0.0'
+
+description = 'OpenCADC VOSpace utility file system library'
+def git_url = 'https://github.com/opencadc/vos'
+
+dependencies {
+    implementation 'org.opencadc:cadc-util:[1.9.0,)'
+    implementation 'org.opencadc:cadc-log:[1.0,)'
```
pdowler, line 29: cadc-log not needed by this lib - when making a new lib, always strip the dependencies down to the minimum

pdowler (changes requested):
You need to add cadc-util-fs to the github workflow to get CI to work.

cadc-util-fs/build.gradle
yeunga: Done.

cadc-util-fs/build.gradle
yeunga: Done

cadc-util-fs/src/main/java/org/opencadc/util/fs/ExtendedFileAttributes.java
```diff
@@ -0,0 +1,126 @@
...
+import java.nio.file.FileSystemException;
+import java.nio.file.Files;
+import java.nio.file.LinkOption;
+import java.nio.file.Path;
+import java.nio.file.attribute.UserDefinedFileAttributeView;
+
+import org.apache.log4j.Logger;
+
+/**
+ * An implementation of the StorageAdapter interface on a file system.
```
pdowler, line 82: This javadoc comment should be revised.

Also, please add javadoc to the two methods; in particular, the set method needs to be documented to say that a null value means delete the attr.

The path and attributeKey values are assumed not null (they never were where this code was used). Now that this is a library, that needs to be well documented and checked (throw IllegalArgumentException if either is null).

pdowler (changes requested):

pdowler (approved):

PR #171, opened 2023-09-07
Re-factor `NodeFault` to allow caller to add specific messages (CADC-12722)

andamian:


cadc-vos-server-alt/src/main/java/org/opencadc/vospace/server/NodeFault.java
```diff
@@ -83,51 +86,62 @@
...
-            return message;
+    public Exception getStatus(String msg) {
+        String exMsg = status;
+        if (StringUtil.hasText(msg)) {
+            exMsg += ": " + msg;
         }
-        if (status != null) {
-            return status.getMessage();
+        switch (status) {
+            case VOS.IVOA_FAULT_PERMISSION_DENIED:
```
pdowler, line 118: permission denied should be AccessControlException

there is standard exception -> http status code mapping in cadc-rest and standard status code -> exception mapping in cadc-util  (HttpTransfer.checkErrors) that is supposed to reverse it

cadc-vos-server-alt/src/main/java/org/opencadc/vospace/server/NodeFault.java
```diff
@@ -83,51 +86,62 @@
...
+            case VOS.CADC_FAULT_UNREADABLE_LINK:
+                return new ResourceNotFoundException(exMsg);
+            case VOS.IVOA_FAULT_DUPLICATE_NODE:
+                return new ResourceAlreadyExistsException(exMsg);
+            case VOS.CADC_FAULT_SERVICE_BUSY:
+                return new TransientException(exMsg);
+            case VOS.CADC_FAULT_NODE_LOCKED:
+                return new ResourceLockedException(exMsg);
+            case VOS.CADC_FAULT_REQUEST_TOO_LARGE:
+                return new InlineContentException(exMsg);
```
pdowler, line 137: I think request too large should be ByteLimitExceededException

pdowler (changes requested):

pdowler (approved):

PR #177, opened 2023-09-15
Set `vault` service state CADC-12718

andamian:


cadc-vos-server-alt/src/main/java/org/opencadc/vospace/server/actions/NodeAction.java
```diff
@@ -128,6 +128,13 @@ public void initAction() throws Exception {
         } catch (Exception oops) {
             throw new RuntimeException("BUG: NodePersistence implementation not found with JNDI key " + jndiNodePersistence, oops);
         }
+
+        checkReadable();
```
pdowler, line 132: I generally prefer having each concrete class implement call the right method at the start of initAction or doAction rather than some reflection-based logic in the base class. Less fragile in the long run and easier to code review changes in the order of things later.

For now, this is probably OK since we have a common initAction here, but I expect it might change before long.

pdowler (approved):

cadc-vos-server-alt/src/main/java/org/opencadc/vospace/server/actions/NodeAction.java
andamian: I thought about this, but as you've said, the common `initAction` action is too simple right now to warrant separate methods that repeat the same code.

PR #178, opened 2023-09-22
CADC-12765 merged in the cavern uid/gid PR

jburke-cadc:
Merged in the uid/gid PR, and updated LocalServiceURI to use cavern.properties.


pdowler (changes requested):
now that I look at this change to get rid of that config file, this seems like a bad idea... I don't know if someone else (INAF maybe) still builds and runs a vospace based on our libs.

What about keep the old constants and add ALT_CONFIG_FILE and ALT_RESOURCE_ID_KEY? Then look for both files and use the one that is found with the associated key?

jburke-cadc:
That seems like a much better idea. Making a minor version specifically for this use case doesn't stop others from unknowingly downloading it and breaking their app. I'll update.

pdowler:
could also then be 1.3.5

cavern/src/main/java/org/opencadc/cavern/CavernInitAction.java
```diff
@@ -86,6 +91,21 @@ public CavernInitAction() {
...
     public void doInit() {
         try {
+            // Check config file
+            CavernConfig cavernConfig = new CavernConfig();
+            MultiValuedProperties props = cavernConfig.getConfig();
+
+            // create root directories for node/files
+            String baseDir = props.getFirstPropertyValue(CavernConfig.FILESYSTEM_BASE_DIR);
+            String subPath = props.getFirstPropertyValue(CavernConfig.FILESYSTEM_SUB_PATH);
+            Path root = Paths.get(baseDir + subPath);
```
pdowler, line 101: make sure there is a `/` between these?

cavern/src/main/java/org/opencadc/cavern/FileSystemNodePersistence.java
```diff
@@ -111,23 +109,18 @@ public class FileSystemNodePersistence implements NodePersistence {
...
     public FileSystemNodePersistence() {
         CavernConfig cavernConfig = new CavernConfig();
         this.config = cavernConfig.getConfig();
-        String rootConfig = config.getFirstPropertyValue(CavernConfig.FILESYSTEM_BASE_DIR);
-        if (rootConfig == null) {
-            throw new RuntimeException("CONFIG: Failed to find " + CavernConfig.FILESYSTEM_BASE_DIR);
-        }
-        this.root = Paths.get(rootConfig);
-        
+        String baseDir = config.getFirstPropertyValue(CavernConfig.FILESYSTEM_BASE_DIR);
```
pdowler, line 117: Why repeat this code? Why not `this.root = config.getNodeRoot();`??

cavern/src/main/java/org/opencadc/cavern/FileSystemNodePersistence.java
```diff
@@ -111,23 +109,18 @@ public class FileSystemNodePersistence implements NodePersistence {
...
+        String subPath = config.getFirstPropertyValue(CavernConfig.FILESYSTEM_SUB_PATH);
+        this.root = Paths.get(baseDir + subPath);
+
         // must be hard coded to this and not set via java system properties
         this.identityManager = new PosixIdentityManager();
-        
+
         LocalAuthority loc = new LocalAuthority();
         // TODO: move constant to cadc-registry
         URI posixMapperID = loc.getServiceURI("http://www.opencadc.org/std/posix#group-mapping-1.0");
```
pdowler, line 126: I have fixed this in my current code to:
```
URI posixMapperID = loc.getServiceURI(Standards.POSIX_GROUPMAP.toASCIIString());
```
but not pushed yet.

cavern/src/test/resources/cadc-registry.properties
```diff
@@ -8,4 +8,4 @@ ca.nrc.cadc.reg.client.RegistryClient.baseURL = https://example.net/reg
 ivo://ivoa.net/std/GMS#search-1.0 = ivo://cadc.nrc.ca/gms
 ivo://ivoa.net/std/UMS#users-0.1 = ivo://cadc.nrc.ca/gms
 
-http://www.opencadc.org/std/posix#group-mapping-1.0 = ivo://cadc.nrc.ca/gms
\ No newline at end of file
+http://www.opencadc.org/std/posix#group-mapping-1.0 = ivo://cadc.nrc.ca/gms
```
pdowler, line 11: again, 0.1

pdowler (changes requested):

cavern/src/main/java/org/opencadc/cavern/CavernInitAction.java
jburke-cadc: Yes, would be a good idea if the base doesn't end with a / and the subPath doesn't start with one.

cavern/src/main/java/org/opencadc/cavern/FileSystemNodePersistence.java
jburke-cadc: Updated my copy.

cavern/src/test/resources/cadc-registry.properties
jburke-cadc: Updated.

pdowler (approved):

PR #186, opened 2023-10-06
 Recursive delete int tests and support for node lock (CADC-12558)

andamian:


cadc-vos-server-alt/src/main/java/org/opencadc/vospace/server/async/JobManager.java
```diff
@@ -0,0 +1,87 @@
...
+
+package org.opencadc.vospace.server.async;
+
+import ca.nrc.cadc.auth.AuthenticationUtil;
+import ca.nrc.cadc.uws.server.JobPersistence;
+import ca.nrc.cadc.uws.server.SimpleJobManager;
+import ca.nrc.cadc.uws.server.impl.PostgresJobPersistence;
+import org.apache.log4j.Logger;
+
+public abstract class JobManager extends SimpleJobManager {
```
pdowler, line 76: delete this class

cadc-vos-server-alt/src/main/java/org/opencadc/vospace/server/async/RecursiveDeleteNodeJobManager.java
```diff
@@ -0,0 +1,101 @@
...
+import ca.nrc.cadc.uws.server.JobExecutor;
+import ca.nrc.cadc.uws.server.JobUpdater;
+import ca.nrc.cadc.uws.server.ThreadPoolExecutor;
+import org.apache.log4j.Logger;
+
+/**
+ *
+ * @author pdowler, majorb, yeunga, adriand
+ */
+public class RecursiveDeleteNodeJobManager extends JobManager {
```
pdowler, line 81: delete this file

cadc-vos-server-alt/src/main/java/org/opencadc/vospace/server/async/RecursiveDeleteNodeRunner.java
```diff
@@ -0,0 +1,358 @@
...
+
+            if (ep == null) {
+                throw new IllegalStateException(
+                        "Could not change the job phase from " + ExecutionPhase.QUEUED
+                                + " to " + ExecutionPhase.EXECUTING);
+            }
+
+            VOSURI nodeURI = null;
+            for (Parameter param : job.getParameterList()) {
+                if (param.getName().equalsIgnoreCase("nodeURI")) {
```
pdowler, line 165: synctrans has a parameter to denote the target node and it's in the VOSpace-2.1 spec - what is the param name there? should use the same if it makes sense

also, container listing uses `uri` param as the "start at this one" so that's also a possibility

cadc-vos-server-alt/src/main/java/org/opencadc/vospace/server/async/RecursiveDeleteNodeRunner.java
```diff
@@ -0,0 +1,358 @@
...
+            if (serverNode instanceof ContainerNode) {
+                try {
+                    deleteContainer((ContainerNode) serverNode, caller);
+                } catch (JobAbortedException ex) {
+                    // nothing to do here
+                }
+            } else {
+                try {
+                    nodePersistence.delete(serverNode);
+                    log.debug("Recursively deleted data node " + nodePath);
```
pdowler, line 199: this debug message will be misleading because this is the non-recursive option

cadc-vos-server-alt/src/main/java/org/opencadc/vospace/server/async/RecursiveDeleteNodeRunner.java
```diff
@@ -0,0 +1,358 @@
...
+            sendError("Unexpected Exception: " + t.getMessage());
+        }
+    }
+
+    private boolean deleteContainer(ContainerNode node, Subject caller) throws Exception {
+
+        boolean errors = false;
+        List<ContainerNode> childContainers = new ArrayList<>();
+        log.debug("Deleting container " + Utils.getPath(node));
+        ResourceIterator<Node> iterator = nodePersistence.iterator(node, null, null);
```
pdowler, line 255: this needs to use `try-with-resources` so that `close()` is automatically and always called

cadc-vos-server-alt/src/main/java/org/opencadc/vospace/server/async/RecursiveDeleteNodeRunner.java
```diff
@@ -0,0 +1,358 @@
...
+        while (iterator.hasNext()) {
+            checkJobPhase();
+            Node child = iterator.next();
+            if (child instanceof ContainerNode) {
+                childContainers.add((ContainerNode) child);
+            } else {
+                try {
+                    if (!vospaceAuthorizer.hasSingleNodeWritePermission(child.parent, caller)) {
+                        throw NodeFault.PermissionDenied.getStatus(Utils.getPath(child));
```
pdowler, line 264: don't use throw and catch (line 269) for logic, just something like
```
if (vospaceAuthorizer.hasSingleNodeWritePermission(child.parent, caller)) {
    // do the delete
} else {
    // track error
}
```

also, that call to vospaceAuthorizer is the same for every iteration so doesn't need to be repeated inside the loop... BUT: does this take into account a locked data node inside a writable container node?

pdowler (changes requested):
should not have  JobManager(s) in async package

cadc-vos-server-alt/src/main/java/org/opencadc/vospace/server/async/RecursiveDeleteNodeRunner.java
andamian: Replaced with `target` is per async endpoint.

cadc-vos-server-alt/src/main/java/org/opencadc/vospace/server/async/RecursiveDeleteNodeRunner.java
andamian: Removed. The idea was to show that the file was deleted as a result of the recursive delete but that it's already illustrated by the name of the file.

cadc-vos-server-alt/src/main/java/org/opencadc/vospace/server/async/RecursiveDeleteNodeRunner.java
andamian: Done. Good point.

cadc-vos-server-alt/src/main/java/org/opencadc/vospace/server/async/RecursiveDeleteNodeRunner.java
andamian: Done. Good catch

pdowler (changes requested):

cadc-vos-server-alt/src/main/java/org/opencadc/vospace/server/async/RecursiveDeleteNodeRunner.java
```diff
@@ -0,0 +1,382 @@
...
+                        nodePersistence.delete(child);
+                        log.debug("Deleted non-container node " + Utils.getPath(child));
+                        deleteCount++;
+                    } else {
+                        log.debug("Unauthorized to delete node " + Utils.getPath(child));
+                        errors = true;
+                        incErrorCount();
+                    }
+                }
+            }
```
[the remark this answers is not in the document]
andamian: Agree.

pdowler (approved):

PR #204, opened 2023-12-01
Add empty properties for detail=min.

at88mph:


pdowler:
empty properties element not required by xsd

PR #208, opened 2023-12-12
CADC-12563 generic PackageRunner classes for cavern and vault

jburke-cadc:


cadc-vos-server/src/main/java/org/opencadc/vospace/server/pkg/VospacePackageRunner.java
```diff
@@ -0,0 +1,379 @@
...
+            // check job is valid
+            log.debug("job id passed in: " + job.getID());
+            JobInfo jobInfo = job.getJobInfo();
+
+            // Get the target list from the job
+            TransferReader tr = new TransferReader();
+            this.packageTransfer = tr.read(jobInfo.getContent(), VOSURI.SCHEME);
+            List<URI> targetList = packageTransfer.getTargets();
+            if (targetList.size() > 1) {
+                this.packageName = "cadc-download-" + job.getID();
```
pdowler, line 163: instead of `cadc-download-` this should use appname (eg cavern or arc or vault), eg `appname + "-download-"` as the base name... then append jobID or name as approp.

cadc-vos-server/src/main/java/org/opencadc/vospace/server/pkg/VospacePackageRunner.java
```diff
@@ -0,0 +1,379 @@
...
+     *
+     * @return List of PackageItem's.
+     */
+    @Override
+    protected Iterator<PackageItem> getItems() {
+        // packageTransfer is populated in initPackage
+        List<URI> targetList = packageTransfer.getTargets();
+        log.debug("target list: " + targetList.toString());
+
+        List<PackageItem> packageItems = new ArrayList<>();
```
pdowler, line 185: This looks like it accumulates the entire list of items (nodes) in memory before returning. That's going to have high latency (failure mode) and consume an arbitrary amount of memory.

The whole implementation of recursively adding content has to happen inside the Iterator<PackageItem> and traversal has to happen while the PackageRunner iterates.

pdowler (changes requested):

cadc-vos-server/src/main/java/org/opencadc/vospace/server/pkg/VospacePackageRunner.java
jburke-cadc: Updated the package name.

cadc-vos-server/src/main/java/org/opencadc/vospace/server/pkg/VospacePackageRunner.java
jburke-cadc: Updated to return an iterator over the PackageItem's.

I left the advance logic in the method. At first used separate methods to refresh the iterators, but it's easier to understand when it's all in one place.

pdowler (changes requested):
iterator advance()

clarify null target nodes (not found) and skip, not break

line 371: the node needs to generate the next PackageItem
line 373: do read permission check on node before nodePersistence.iterator( ) call, skip if denied

use `return` instead of `break`

cadc-test-vos/src/main/java/org/opencadc/conformance/vos/VOSTest.java
```diff
@@ -124,6 +124,7 @@ public abstract class VOSTest {
     public final URL synctransServiceURL;
     public final URL transferServiceURL;
     public final URL recursiveDeleteServiceURL;
+    public final URL pkgServiceURL;
```
pdowler, line 127: Is there a good reason that subclass-specific endpoints are accumulating here? For this one in particular, there are no pkg-specific methods in the base class... I would think the supporet methods here would only need the nodes endpoint.

Probably something I didn't pay attention to from earlier work,  but it looks more and more like poor style.

pdowler (changes requested):

cadc-vos-server/DEVREADME.md
```diff
@@ -0,0 +1,31 @@
+A transfer to download a package (a compressed collection of resources) uses a different sequence of job phases
```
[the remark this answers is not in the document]
jburke-cadc: Renamed to PackageDownload.md

cadc-vos-server/DEVREADME.md
```diff
@@ -0,0 +1,31 @@
...
+a Transfer with a Package View, it puts the job into the suspended phase. In UWS the SUSPENDED phase indicates
+that the job is suspending awaiting further processing. After the TransferRunner suspends the job, 
+it redirects to the PackageRunner to run the job, return the package, and complete the job.
+
+The sequence for a synchronous package transfer:
+
+- POST a transfer document with a package view to the /<service>/synctrans endpoint.
+- A new job is created in the PENDING phase, then a redirect is issued to /<service>/synctrans/<jobID>/run.
+- GET the redirect, which set the job phase from PENDING to QUEUED, and runs the TransferRunner.
+- The TransferRunner validates the job and transfer, finds a package view in the transfer,
```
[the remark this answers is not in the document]
jburke-cadc: I was expecting the phase change to be one of the first things the TransferRunner did, and went looking everywhere else but the end of the run trying to find it.

cadc-test-vos/src/main/java/org/opencadc/conformance/vos/VOSTest.java
jburke-cadc: A couple are common to multiple tests, nodes and synctrans. Subsequent tests probably just followed the same pattern. I forgot to remove pkgServiceURL, it's not used. Moved the others to the tests where they are used.

pdowler (approved):

PR #213, opened 2024-01-17
link node tests and cavern limit handling

pdowler:
cavern will ignore limit param rather than reject


pdowler:
commits are part of #214

PR #214, opened 2024-01-18
Fixed support for LinkNodes (CADC-13048)

andamian:


cadc-vos-server/src/main/java/org/opencadc/vospace/server/PathResolver.java
```diff
@@ -105,9 +105,8 @@ public class PathResolver {
     private final VOSpaceAuthorizer voSpaceAuthorizer;
     private final boolean resolveLinks;
     
-    private List<String> visitedPaths = new ArrayList<>();
     private int visitLimit = 20;
-    private int visitCount = 0;
+
```
pdowler, line 111: The class javadoc should not mention "resolving" and there really needs to be a javadoc here for the constructor to explain exactly how `resolveLinks` effects `getNode(...)` -- for our own benefit!

I would change the 3rd arg to `resolvelink` (singular since it only effects the last name in the target path).

Optional: since most of the code (actions) use resolveLink = false, it might be a good idea to have these ctors instead:
```
PathResolver(NodePersistence nodePersistence, VOSpaceAuthorizer voSpaceAuthorizer) {
   this(nodePersistence, voSpaceAuthorizer, false);
}

PathResolver(NodePersistence nodePersistence, VOSpaceAuthorizer voSpaceAuthorizer, boolean resolveLink) ...
```
use the shorter one in all the actions and the special case where explicitly needed in transfer code... but maybe that makes things more obscure?

cadc-vos-server/src/main/java/org/opencadc/vospace/server/PathResolver.java
```diff
@@ -182,8 +186,8 @@ public Node getNode(String nodePath) throws Exception {
                 ret = child;
             }
         }
-        
-        log.debug("return node: " + Utils.getPath(ret));
+
+        log.debug("return node: " + (( ret != null) ? Utils.getPath(ret) : null));
```
pdowler, line 190: Is there another way to get here with ret==null? There are already immediate `return null;` up at line 146 and 179

cavern/src/main/java/org/opencadc/cavern/files/CavernURLGenerator.java
```diff
@@ -148,20 +149,31 @@ public List<Protocol> getEndpoints(VOSURI target, Transfer transfer, Job job, Li
...
-            ContainerNode parent = (ContainerNode) n;
-            Node node = nodePersistence.get(parent, target.getName());
+            PathResolver pr = new PathResolver(nodePersistence, authorizer, true);
+            Node node = pr.getNode(target.getPath());
+            ContainerNode parent;
+            String name;
+            if (node == null) {
+                // maybe a DataNode that needs to be created later. Check the parent (Note: a bit repetitive to check
+                // the entire path again)
+                Node tmp = pr.getNode(target.getParent());
```
pdowler, line 159: I don't really like this approach of doing the path twice. Isn't it sufficient to leave it as before and maybe just check before casting? 

Conceptually,  I think it is usually the case that the data node (file) should not exist (the create side effect farther upstream in transfer negotiation is kind of a bug I'd like to fix) unless there was an explicit create DataNode (or a file is being replaced). When I fix that, this code will be quite sub-optimal.

pdowler (changes requested):

cadc-vos-server/src/main/java/org/opencadc/vospace/server/PathResolver.java
andamian: Rename the argument and added docs.

cadc-vos-server/src/main/java/org/opencadc/vospace/server/PathResolver.java
andamian: Sometimes when it comes out of lane 85. I've already run into it. Although now with the alternative implementation in `CavernURLGenerator` might not be the case, it is still possible.

cavern/src/main/java/org/opencadc/cavern/files/CavernURLGenerator.java
andamian: Re-worked to avoid redundant calls. I was thinking to to it this way but didn't realize `validateTargetURI` was available and I didn't want to do it just for this.

cadc-vos-server/src/main/java/org/opencadc/vospace/server/transfers/PushToVOSpaceNegotiation.java
```diff
@@ -112,30 +111,16 @@ public void doAction()
...
-            //Node node = resolveNodeForWrite(authorizer, nodePersistence, target, DataNode.class, true, true, true);
-            Node node = nodePersistence.get(parent, target.getName());
-            log.debug("target node: " + target + " -> " + node);
-            
-            DataNode dn = null;
-            if (node == null) {
-                if (!authorizer.hasSingleNodeWritePermission(parent, caller)) {
-                    throw NodeFault.PermissionDenied.getStatus(target.getParentURI().getURI().toASCIIString());
-                }
+            if (resolved.child == null) {
```
pdowler, line 120: this is now missing two checks:
`resolved.parent != null` 
`hasSingleNodeWritePermission(resolved.parent, caller)`

cavern/src/main/java/org/opencadc/cavern/files/CavernURLGenerator.java
```diff
@@ -146,28 +144,26 @@ public List<Protocol> getEndpoints(VOSURI target, Transfer transfer, Job job, Li
...
-            if (Direction.pushToVoSpace.equals(dir) && node == null) {
-                // create new data node?? this currently does not happen because the library
-                // creates the DataNode
-                ret = handleDataNode(parent, target.getName(), transfer, currentSubject);
-            } else if (node instanceof DataNode) {
-                ret = handleDataNode(parent, target.getName(), transfer, currentSubject);
-            } else if (node instanceof ContainerNode) {
-                ret = handleContainerMount(target.getPath(), transfer, currentSubject);
+            PathResolver pr = new PathResolver(nodePersistence, authorizer, true);
+            PathResolver.Resolved resolved = pr.resolveNode(target);
```
pdowler, line 150: I thought the idea was that this should be able to use `Node node = pr.get(target path)` now? The spec requires that the DataNode is created during transfer negotiation. 

Admittedly, "transfer negotiation" spans multiple requests and this code (URL generation) happens in a later request but I think it was clear that we create the target DataNode in the first step (TransferRunner + PushToVOSpaceNegotiation) because we want to do all the checking in the first call rather than later... otoh, it is feasible that the new DataNode gets deleted by some other actor between when it is created in TransferRunner and then this code runs in TransferDetailsServlet, which could be an arbitrary amount of time later. So this code has to check that the DataNode still exists but should fail if it does not.

Given that we already noted that the DataNode create in PushToVOSpaceNegotiation must be consistent with whatever CreateNodeAction does, having another possible create in here is a bad idea.

pdowler (changes requested):

cadc-vos-server/src/main/java/org/opencadc/vospace/server/transfers/PushToVOSpaceNegotiation.java
andamian: It's `resolveNode` responsibility to return a valid `parent` and `childName` when `child` not found. Good point about `hasSingleNodeWritePermission(resolved.parent, caller)`. Fixed

cadc-vos-server/src/main/java/org/opencadc/vospace/server/transfers/PullFromVOSpaceNegotiation.java
```diff
@@ -152,12 +152,12 @@ public void doAction() throws Exception {
...
             // careful to capture a link to data node so we can get the right filename in the transfer
-            PathResolver resolver1 = new PathResolver(nodePersistence, authorizer, false);
-            Node apparentNode = resolver1.getNode(target.getPath());
+            PathResolver resolver1 = new PathResolver(nodePersistence, authorizer);
+            Node apparentNode = resolver1.getNode(target.getPath(), false);
             Node actualNode = apparentNode;
             if (apparentNode instanceof LinkNode) {
-                PathResolver resolver2 = new PathResolver(nodePersistence, authorizer, true);
-                actualNode = resolver2.getNode(target.getPath());
+                PathResolver resolver2 = new PathResolver(nodePersistence, authorizer);
```
pdowler, line 159: no longer need a new PathResolver here

cadc-vos-server/src/main/java/org/opencadc/vospace/server/transfers/InternalTransferAction.java
```diff
@@ -138,15 +138,16 @@ public void doAction() throws Exception {
             log.debug("checking move permissions: " + srcURI + " -> " + destURI);
 
             // resolve the links to containers in the path so we get the actual node
-            PathResolver res = new PathResolver(nodePersistence, authorizer, true);
-            Node srcNode = res.getNode(srcURI.getPath());
+            PathResolver res = new PathResolver(nodePersistence, authorizer);
+            Node srcNode = res.getNode(srcURI.getPath(), true);
```
pdowler, line 142: I'm not sure resolveLeafLink==true is right here. It means you don't move a LinkNode and end up moving the target of the link node, which also breaks the link node.... it was probably always a bug that was hiding due to the ambiguous meaning of "resolve"

at least the TODO comment below is no longer a concern and can be deleted :-)

pdowler (changes requested):

PR #216, opened 2024-02-02
CADC-12859 add package support to vault

jburke-cadc:


pdowler:
The dependency on cadc-pkg-runner should not be updated to require latest when it's only a toString() method to help debugging... that will allow this to build. If tyou reaklly need that toString method, that's probably a bug because you should rarely/never rely on toString() for specific behaviour

pdowler:
nevermind: cadc-pkg-server published

cadc-vos-server/src/main/java/org/opencadc/vospace/server/actions/CreateNodeAction.java
```diff
@@ -199,7 +199,7 @@ public void doAction() throws Exception {
         
         // output modified node
         NodeWriter nodeWriter = getNodeWriter();
-        syncOutput.setCode(200); // Created??
+        syncOutput.setCode(201); // Created??
```
pdowler, line 202: comment should be removed since it was there to note this exact thing (201 means "Created" in HTTP)

cadc-vos-server/src/test/java/org/opencadc/vospace/server/VospacePackageRunnerTest.java
```diff
@@ -0,0 +1,130 @@
...
+
+        // should fail
+        URI node = URI.create("vos://opencadc.org~vault/pkg/foo/link");
+        URI target = URI.create("http://example.com/target");
+        String relativePath = "/pkg/foo/";
+        boolean result = VospacePackageRunner.isTargetInPackage(node, target, relativePath);
+        Assert.assertFalse(result);
+
+        node = URI.create("vos://opencadc.org~vault/pkg/foo/link");
+        target = URI.create("vos://example.com~vault/pkg/target");
```
pdowler, line 99: this target is outside for two reasons so doesn't necessarily test both

cadc-vos-server/src/main/java/org/opencadc/vospace/server/pkg/VospacePackageRunner.java
```diff
@@ -122,6 +125,7 @@ public class VospacePackageRunner extends PackageRunner {
     private PathResolver pathResolver;
     private List<URI> targetList;
     private String appName;
+    private String baseNodeUri;
```
pdowler, line 128: String type and "Uri" in the name is a bad sign here

cadc-vos-server/src/main/java/org/opencadc/vospace/server/pkg/VospacePackageRunner.java
```diff
@@ -142,11 +146,51 @@ public void setAppName(String appName) {
...
     }
 
+    /**
+     *
+     * @param node
+     * @param target
+     * @param relativePath
+     * @return
+     */
+    public static boolean isTargetInPackage(URI node, URI target, String relativePath) {
```
pdowler, line 162: no reason for this to be public, just package access for test is fine
don't need javadoc, especially one that doesn't say anything useful, but a plain comment with explanation would be nice; I can't tell what the args mean

I am surprised that you need 3 args for this... and that there is not a LinkNode in the arg list

cadc-vos-server/src/main/java/org/opencadc/vospace/server/pkg/VospacePackageRunner.java
```diff
@@ -237,35 +281,40 @@ protected URL getURL(VOSURI nodeURI)
...
-        log.debug("ContainerNode path: " + nodePath);
-
-        return new PackageItem(nodePath);
+    protected PackageItem getDirectoryPackageItem(ContainerNode node, String relativePath) {
+        return new PackageItem(relativePath);
     }
 
     // Get a PackageItem for a file from the given DataNode.
-    protected PackageItem getFilePackageItem(DataNode node)
+    protected PackageItem getFilePackageItem(DataNode node, String relativePath)
```
pdowler, line 289: in getDirectoryPackageItem the ContainerNode arg is ignored and only the relativePath is used, but here in getFilePackageItem it uses both... inconsistent usage style leads to bugs.

It looks like the second arg should be named `parentPath` and probably swapped to be first. In that case getDirectoryPackageItem is the one that's wrong (and maybe how it is called is wrong -- 2 bugs that counter each other?)

cadc-vos-server/src/main/java/org/opencadc/vospace/server/pkg/VospacePackageRunner.java
```diff
@@ -237,35 +281,40 @@ protected URL getURL(VOSURI nodeURI)
...
         log.debug("LinkNode path: " + nodePath);
 
-        URI targetURI = node.getTarget();
+        // check that the link target is in the package
+        // return a null PackageItem if it is out of the package
         VOSURI nodeURI = new VOSURI(this.resourceID, nodePath);
-        String linkPath = nodeURI.getPath();
-        log.debug("LinkNode target path: " + linkPath);
+        URI targetURI = node.getTarget();
+        boolean inPackage = isTargetInPackage(nodeURI.getURI(), targetURI, relativePath);
```
pdowler, line 308: Since this uses the "early return" strategy (I prefer that), it would be nicer if the utility method was something like 
```
boolean isExternal(LinkNode link, String packageBasePath)
```
so the usage is positive, like 
```
if (externalLink) {
   return null;
}
```
aside: note committed to `isExternal` because it might be too vague, maybe `isOutsidePackage`???

Secondarily, if there are multiple targets in the package job, that means there are multiple basePkgPath(s) which define internal vs external. I don't think we discussed that explicitly but as soon as I tried to write the above method signature is became obvious.

cadc-vos-server/src/main/java/org/opencadc/vospace/server/pkg/VospacePackageRunner.java
```diff
@@ -506,9 +557,20 @@ private PackageItem doChildNode(Node child) {
...
             }
+            log.debug("doChildNode().end");
             return packageItem;
         }
 
     }
 
+    public class RelativeNode {
+        public ContainerNode node;
+        public String path;
```
pdowler, line 568: if this is `parentPath` please make it clear

and if so reverse the order of ctor args: if it is parent, child (to be appended in some logical sense) it should be in that order

cavern/VERSION
```diff
@@ -1,6 +1,6 @@
 ## deployable containers have a semantic and build tag
 # semantic version tag: major.minor
 # build version tag: timestamp
-VER=0.6.3
+VER=0.6.4
```
pdowler, line 4: currently published cavern image is 0.6.2

pdowler (changes requested):

cadc-test-vos/build.gradle
```diff
@@ -16,7 +16,7 @@ sourceCompatibility = 1.8
 
 group = 'org.opencadc'
 
-version = '2.1.6'
+version = '2.1.7'
```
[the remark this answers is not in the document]
jburke-cadc: Ah okay, I was looking at the latest GitHub build version, reverted.

cadc-vos-server/build.gradle
```diff
@@ -16,7 +16,7 @@ sourceCompatibility = 1.8
 
 group = 'org.opencadc'
 
-version = '2.0.9'
+version = '2.0.10'
```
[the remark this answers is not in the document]
jburke-cadc: Same as above, checking the wrong place, reverted.

cadc-vos-server/src/main/java/org/opencadc/vospace/server/actions/CreateNodeAction.java
jburke-cadc: Removed the comment.

cadc-vos-server/src/test/java/org/opencadc/vospace/server/VospacePackageRunnerTest.java
jburke-cadc: Not seeing the second reason.
In the context where isTargetInPackage is called, the relative path is the relative path to the link node parent. The relative path and the link node name are used to compose the relative path to the node, which is passed to the PackageItem constructor. 
Although it might be better if it was the relative path to the link node in isTargetInPackage, and not the parent. Might change that.

cadc-vos-server/src/main/java/org/opencadc/vospace/server/pkg/VospacePackageRunner.java
jburke-cadc: I usually use xxxURI if the type is URI, and xxxUri for a string that is a URI, but of course no one else would know that. Updated the name.

jburke-cadc: And as it turns out, not used. I think that was the first pass at figuring out if a target was in the package tree. Removed.

cadc-vos-server/src/main/java/org/opencadc/vospace/server/pkg/VospacePackageRunner.java
jburke-cadc: I think I already had the node and target URI at one time, seemed proper to use them as args, but that was the first pass again. Does make more sense to pass in the LinkNode. Updated.

jburke-cadc: Forgot to complete the javadoc, updated.

cadc-vos-server/src/main/java/org/opencadc/vospace/server/pkg/VospacePackageRunner.java
jburke-cadc: The relative path is consistent between all 3, so it should be the first arg. For data and link nodes the relative path is the path to the parent node so calling it the parent node makes better sense. For container nodes, it's the path to the node itself. Code comments should make that more clear.

cadc-vos-server/src/main/java/org/opencadc/vospace/server/pkg/VospacePackageRunner.java
jburke-cadc: `isExternalTarget`? Yes there can be multiple targets, and the relative path is the path to the target for the node.

cadc-vos-server/src/main/java/org/opencadc/vospace/server/pkg/VospacePackageRunner.java
jburke-cadc: It's a class for a container node, and the path from the target to the node.
The relative path use is confusing because it's not consistent. It means something different for a container node compared to a data or link node. And as I think about it, I don't think it needs to. It could consistently be the path from the target to the parent node, for all nodes.
See if I can make that change.

cavern/VERSION
jburke-cadc: One last time hopefully, updated.

cadc-vos-server/src/test/java/org/opencadc/vospace/server/VospacePackageRunnerTest.java
pdowler: outside the package (path) or outside the service (external link)

I am interpreting the test as `node` is the root of the package and `target` is the target of the link. Maybe part of the issue is that `node` might as well be `n` or `obj` because it doesn't convey any meaning at all... and it's not even a Node, it's a URI. Which node is it?

1. the target has a different authority
2. the child path is not under the node path

iirc, a different test checks "different authority" ... well I thought so, but all the node and target have different authority so now I really don't know what's being tested here

cadc-vos-server/src/main/java/org/opencadc/vospace/server/pkg/VospacePackageRunner.java
pdowler: by multiple targets I meant multiple input URIs to the pkg job, eg
```
\/a
\/b
```
In that case, a link node somewhere under `\/a` that points under `\/b` *should* be included in  the package. So the package as a whole has multiple starting paths, not just one, and `isExternalTarget` needs to take that into account.

jburke-cadc: So `isExternalTarget` has to check if the link target is under all the input targets. Okay, I'll see how that can be done.

cadc-vos-server/src/main/java/org/opencadc/vospace/server/pkg/VospacePackageRunner.java
pdowler: why is the meaning of the relative path different for container nodes? it seems like a sign that something else is not right as well... it might work, but will look like a bug and be easy to break if inconsistent

pdowler (approved):

PR #222, opened 2024-02-29
Support for allocation owner permissions on their tree (CADC-13241)

andamian:


cavern/src/test/resources/cadc-registry.properties
```diff
@@ -1,6 +1,6 @@
 # configure RegistryClient bootstrap
 ## BUG / HACK: this gets pulled into the intTest target so needs a legit value
-ca.nrc.cadc.reg.client.RegistryClient.baseURL = https://haproxy.cadc.dao.nrc.ca/reg/
+ca.nrc.cadc.reg.client.RegistryClient.baseURL = https://localhost.cadc.dao.nrc.ca/reg/
```
pdowler, line 3: ugh. this test setup was a hack and this solution is no better or worse than before. I will try to fix the test setup separately.

pdowler (approved):

cavern/src/test/resources/cadc-registry.properties
andamian: Sorry. That is my local change but it "leaks" from time to time. I've fixed it and then then I broke it again right after...

PR #223, opened 2024-03-05
CADC-12927 fix node permission inconsistencies

jburke-cadc:


cadc-vos-server/src/main/java/org/opencadc/vospace/server/actions/GetNodeAction.java
```diff
@@ -147,65 +147,68 @@ public void doAction() throws Exception {
         
         if (serverNode instanceof ContainerNode) {
             ContainerNode node = (ContainerNode) serverNode;
-
             log.debug("node: " + node);
 
-            // TBD: sorting parameters
```
pdowler, line 153: Why all these deletes? This diff is impossible to grok 

If I read this correctly, the rejection of unsupported operations are indented some more so not always checked... rejection should always happen as soon as possible and not conditionally.

cavern/src/intTest/java/org/opencadc/cavern/TransferTest.java
```diff
@@ -95,15 +107,20 @@ public class TransferTest extends org.opencadc.conformance.vos.TransferTest {
...
     static {
         Log4jInit.setLevel("org.opencadc.conformance.vos", Level.INFO);
-        
         Log4jInit.setLevel("org.opencadc.vospace", Level.INFO);
         Log4jInit.setLevel("org.opencadc.cavern", Level.INFO);
     }
     
     public TransferTest() {
         super(Constants.RESOURCE_ID, Constants.TEST_CERT);
+
```
pdowler, line 116: check cavern NodesTest setup and move things to Constants for consistency

pdowler (changes requested):

cadc-vos-server/src/main/java/org/opencadc/vospace/server/actions/GetNodeAction.java
jburke-cadc: The code actually changed was minimal. Didn't think to look at the diff. Yes in indentation makes it hard to read. A couple of blocks of code were moved as well which also doesn't help.

The 'if' for the indentation is checking if the caller has permission to list the children, and if they do, do the following that gives the node a iterator over its child nodes. If they don't have permission, then skip getting the child iterator. After that, write the node.

cavern/src/intTest/java/org/opencadc/cavern/TransferTest.java
jburke-cadc: Updated.

cavern/src/intTest/java/org/opencadc/cavern/TransferTest.java
```diff
@@ -173,4 +183,108 @@ public void testPreauthToken() {
...
+     * Test that a user can update a data node they do not own, but have write permission to the node.
+     *
+     * The authSubject user is the owner of the parent ContainerNode and child DataNode.
+     * The altSubject user is a member of the altGroup group.
+     * The altGroup by default does not have write access to the parent or child nodes.
+     *
+     * The altGroup group is given write access to the child node, which should
+     * allow the altSubject to write to the child node.
+     */
+    @Test
```
pdowler, line 197: ok, so this method is here but the "enable" is in the test suite? I assume the rework is incomplete...

pdowler (changes requested):

PR #225, opened 2024-03-07
CADC-12927. fix node permission inconsistencies

jburke-cadc:


cadc-test-vos/src/main/java/org/opencadc/conformance/vos/NodesTest.java
```diff
@@ -965,28 +965,47 @@ public void testPermissions() throws Exception {
         String childPath = parentName + "/" + childName;
         final VOSURI childURI = getVOSURI(childPath);
         final URL childURL = getNodeURL(nodesServiceURL, childPath);
+
```
pdowler, line 968: test setup:

/int-tests/testPermissions
/int-tests/testPermissions/testGroupUser
/int-tests/testPermissions/testListUser

The point of naming it `testPermissions` is so only this method does the setup. it looks like you made a new child node (testListUser) instead of using the one that was there... the names make it sound like it is intentional but I can't grok what for. The cleanup only removes 2 of the 3 nodes.

What is "testListUser" for and why is it named that way?

Why do you need more than /int-tests/testPermissions/x where `x` can be any kind of node?

cadc-test-vos/src/main/java/org/opencadc/conformance/vos/FilesTest.java
```diff
@@ -271,6 +285,59 @@ public void emptyFileTest() throws Exception {
...
+            testNode.isPublic = false;
+            log.debug("testNode: " + nodeURL);
+
+            // Cleanup leftover nodes
+            delete(nodeURL, false);
+
+            // PUT the nodes
+            put(nodeURL, nodeURI, testNode);
+
+            // user without permissions should be able to read the node
```
pdowler, line 307: This is FilesTest so it is testing the /files endpoint; all you have here is re-testing some /nodes behaviour

line 307: user with permission on parent can read the node metadata; this is either covered by NodesTest already or really trivial test setup verification (basically, is /int-tests public?)

line 315: why are you testing updating node props in the FilesTest? irrelevant

where is: user with read permission on parent but not DataNode should not be able to read the bytes (403)?

where is: set DataNode.isPublic = true and repeat previous, now OK (200)?

pdowler (changes requested):
I have not looked at PackageTest and TransferTest yet, but reconsider them once you address NodesTest and FilesTest

cadc-test-vos/build.gradle
```diff
@@ -16,7 +16,7 @@ sourceCompatibility = 1.8
 
 group = 'org.opencadc'
 
-version = '2.1.8'
+version = '2.1.9'
```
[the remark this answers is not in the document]
jburke-cadc: Thought I had used the published script, but I guess not. Updated.

cadc-vos-server/build.gradle
```diff
@@ -16,7 +16,7 @@ sourceCompatibility = 1.8
 
 group = 'org.opencadc'
 
-version = '2.0.11'
+version = '2.0.12'
```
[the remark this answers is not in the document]
jburke-cadc: Same as above.

cadc-test-vos/src/main/java/org/opencadc/conformance/vos/NodesTest.java
jburke-cadc: The test creates the parent container node as the test user.
Gets the parent node as a second test user. Used to fail, now it succeeds.
Writes the child node to the parent as a second test user, which fails.
Gives the second test user write permissions to the parent node.
Writes the child to the parent as the second test user, which now succeeds.
Updates the child quota property as the test user.
Gets the child as the test user, which succeeds. Not sure what that is testing.

So the existing child node is owned by the second test user. 

Getting the parent as a user without read access to the parent should return the parent, because it's in the public int-test node, but not the child.
Given the existing child is put by the second test user, I would need a third test user to test with the existing child. Instead I added a second child written by the test user. The testListUser name is because the second test user should not be able to list the children of the parent.

I guess I could change the ownership of the existing child instead, and clear the read/write groups if needed. That seems overly complicated, and makes this test even harder to understand. Which maybe is a clue I should've written a separate test instead of adding on to this one?

I don't like any of the names in this test. The parent is referred to as parentName, testNode, or nodeURL, which I find confusing. And I don't know why a node has User in the name. But it makes sense to whomever wrote it, so I just go with the style, instead of rewriting it so that it makes better sense to me.

cadc-test-vos/src/main/java/org/opencadc/conformance/vos/FilesTest.java
jburke-cadc: In a rush to finish before I had to leave, and just getting it done instead of paying attention to what I was doing. No excuse for that. Rewrote the test the proper way.

cadc-test-vos/src/main/java/org/opencadc/conformance/vos/FilesTest.java
```diff
@@ -271,6 +279,100 @@ public void emptyFileTest() throws Exception {
...
+            // cleanup
+            delete(nodeURL, false);
+
+            // PUT the node
+            put(nodeURL, nodeURI, testNode);
+
+            // Create a pullFromVoSpace Transfer
+            Transfer transfer = new Transfer(nodeURI.getURI(), Direction.pullFromVoSpace);
+            transfer.version = VOS.VOSPACE_21;
+            Protocol protocol = new Protocol(VOS.PROTOCOL_HTTPS_GET);
```
pdowler, line 307: minor point: you should always add an anon Protocol for systems that use embedded token (like vault)

main point: this is the FilesTest and you are creating a DataNode and doing transfer negotiation; that belongs in TransferTest

This test: HEAD (OK) and GET (DENIED) on the /files endpoint to test GetAction

cadc-test-vos/src/main/java/org/opencadc/conformance/vos/NodesTest.java
pdowler: The only thing that changed is that some read requests are now allowed that used to be rejected.
So why test write permission is rejected? There are tests for that already. Anything extra just obscures the intent of the test. So again, 

**Why do you need more than /int-tests/testPermissions/x where x can be any kind of node?**

pdowler (changes requested):

pdowler (approved):

PR #235, opened 2024-04-24
cadc-vos-server: catch exceptions and close childIterator

pdowler:
prevents potential resource leak in NodePersistence implementation


pdowler:
rework after verbal code review/discussion

PR #240, opened 2024-05-14
Quota plugin configurable

at88mph:
Some code cleanup of unused code as well.

### DIFF
--- a/cavern/README.md
+++ b/cavern/README.md
@@ -57,6 +57,12 @@ org.opencadc.cavern.resourceID = ivo://{authority}/{name}
 # (optional) identify which container nodes are allocations
 org.opencadc.cavern.allocationParent = {top level node}
 
+# (optional) provide a class implementing the org.opencadc.cavern.nodes.QuotaPlugin interface to control Quotas
+# for users.
+# Optional, but the default of NoQuotaPlugin will get used if not specified, and users will have free reign.
+#
+org.opencadc.cavern.nodes.QuotaPlugin = {plugin implementation class}
+
 # base directory for cavern files
 org.opencadc.cavern.filesystem.baseDir = {persistent data directory in container}
 org.opencadc.cavern.filesystem.subPath = {relative path to the node/file content that could be mounted in other containers}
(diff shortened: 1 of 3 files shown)


cavern/README.md
```diff
@@ -57,6 +57,12 @@ org.opencadc.cavern.resourceID = ivo://{authority}/{name}
 # (optional) identify which container nodes are allocations
 org.opencadc.cavern.allocationParent = {top level node}
 
+# (optional) provide a class implementing the org.opencadc.cavern.nodes.QuotaPlugin interface to control Quotas
+# for users.
+# Optional, but the default of NoQuotaPlugin will get used if not specified, and users will have free reign.
+#
```
pdowler, line 63: need to document the choices in the text below

In fenwick and tantar we had plugins that are internal (artifactSelector and ResolutionPolicy respectively) which we started out putting fully qualified class names and then changed to a simple set of terms and mapped that to classes internally. I think for internally implemented (which I expect the quota plugins will be) that's slightly nicer and it gets rid of really dynamic plugin loading.

I'd probably follow the style in tantar.

cavern/src/main/java/org/opencadc/cavern/CavernConfig.java
```diff
@@ -220,7 +221,68 @@ public Subject getRootOwner() {
...
+     *
+     * @param <T>             Class type of the instantiated class
+     * @param implementationClassName   Class name to create
+     * @param constructorArgs The constructor arguments
+     * @return configured implementation of the interface
+     *
+     * @throws IllegalStateException if an instance cannot be created
+     */
+    @SuppressWarnings("unchecked")
+    public static <T> T loadPlugin(final String implementationClassName, final Object... constructorArgs)
```
pdowler, line 256: might not need this if using a hard coded name -> impl mapping

cavern/src/main/java/org/opencadc/cavern/nodes/FileSystemNodePersistence.java
```diff
@@ -183,15 +182,16 @@ public FileSystemNodePersistence() {
...
             try {
                 URL baseURL = posixMapperID.toURL();
-                this.posixMapper = new MyPosixMapperClient(baseURL);
+                posixMapper = new MyPosixMapperClient(baseURL);
             } catch (MalformedURLException ex) {
                 throw new InvalidConfigException("invalid " + Standards.POSIX_GROUPMAP.toASCIIString() + " base URL: " + posixMapperID, ex);
             }
         } else {
-            this.posixMapper = new MyPosixMapperClient(posixMapperID);
+            posixMapper = new MyPosixMapperClient(posixMapperID);
```
pdowler, line 194: prefer to keep the `this.` prefix on the left hand side when assigning to member vars; it stands out when a bit better in a non-syntax highlighting environment

pdowler (changes requested):

cavern/src/main/java/org/opencadc/cavern/nodes/FileSystemNodePersistence.java
at88mph: The `posixMapper` variable was moved into a local variable as it was only used in creating the `GroupCache` object.  It wasn't used as a Class member.

pdowler: ah, ok. makes sense.

pdowler (approved):

PR #241, opened 2024-05-16
Quota plugin cephfs

at88mph:
Minor cleanup
Added quota plugin implementation for CephFS
Yet untested, but here for draft review


cadc-util-fs/src/main/java/org/opencadc/util/fs/PathUtil.java
```diff
@@ -0,0 +1,147 @@
...
+import org.apache.log4j.Logger;
+
+import java.io.IOException;
+import java.nio.file.Path;
+import java.nio.file.Paths;
+
+/**
+ * Utility class to work with Paths.
+ */
+public class PathUtil {
```
pdowler, line 81: this library is simply for set/get of attributes and ACLs; I do not think this method should exist (and hence this util class)

over in the cavern code, there is no scanning for a quota attribute; it is an attribute of a single node and cavern will never scan for it: it will simply be a special property of a single node in a special namespace

cavern/src/main/java/org/opencadc/cavern/nodes/CephFSQuotaPlugin.java
```diff
@@ -0,0 +1,155 @@
...
+     */
+    @Override
+    public Long getQuota(Path directory) {
+        LOGGER.debug("getQuota: " + directory);
+        if (directory == null) {
+            throw new IllegalArgumentException("Cannot check quota of null directory");
+        }
+
+        try {
+            final Long quotaInBytes = PathUtil.scanQuotaFromPath(directory, 0, CephFSQuotaPlugin.QUOTA_ATTR_KEY,
```
pdowler, line 127: This is wrong. the quota is a property of a specific node and this should only check the argument Path.

Most calls will return null except when `directory` has the ceph quota prop

pdowler (changes requested):

cadc-util-fs/src/main/java/org/opencadc/util/fs/PathUtil.java
at88mph: Ah, I was under the impression that the get quota would need to do this.  The implication then is that the `Path` being passed into the `QuotaPlugin` is the intended `Path`.  Will change.

cavern/src/main/java/org/opencadc/cavern/nodes/CephFSQuotaPlugin.java
at88mph: Understood.  I had thought the patch checking would translate into this property checking as well.  Done.

cadc-util-fs/src/main/java/org/opencadc/util/fs/PathUtil.java
at88mph: Done.

cavern/src/main/java/org/opencadc/cavern/nodes/CephFSQuotaPlugin.java
```diff
@@ -0,0 +1,158 @@
...
+        LOGGER.debug("getBytesUsed: " + directory);
+        if (directory == null) {
+            throw new IllegalArgumentException("Cannot get size of null directory");
+        }
+
+        try {
+            final String recursiveBytesUsedValue =
+                    ExtendedFileAttributes.getFileAttribute(directory, CephFSQuotaPlugin.RECURSIVE_SIZE_ATTR_KEY,
+                                                            CephFSQuotaPlugin.NAMESPACE);
+            final long recursiveBytesUsed = Long.parseLong(recursiveBytesUsedValue);
```
pdowler, line 101: this is in principle susceptible to NullPointerException (if attr is null for some reason) and NumberFormatException (unlikely).. both runtime so would cause failure.


I think you are right that "warn" and return null is an acceptable behaviour here, so I'd just add those to the catch.

cavern/src/main/java/org/opencadc/cavern/nodes/CephFSQuotaPlugin.java
```diff
@@ -0,0 +1,158 @@
...
+        try {
+            final String quotaAttributeValue = ExtendedFileAttributes.getFileAttribute(directory,
+                                                                                       CephFSQuotaPlugin.QUOTA_ATTR_KEY,
+                                                                                       CephFSQuotaPlugin.NAMESPACE);
+            final Long quotaInBytes = StringUtil.hasText(quotaAttributeValue)
+                                      ? Long.parseLong(quotaAttributeValue) : null;
+            LOGGER.debug("getQuota: " + directory + ": OK");
+
+            return quotaInBytes;
+        } catch (IOException ioException) {
```
pdowler, line 135: NPE is covered but NumberFormatException is still possible (in principle)... and the message is not quite right now as it isn't "unable to find a folder" but a failed to read filesystem.

Here I am less sure about "warn" and return null. If deployer thinks they have configured quotas then a fail here should probably be a fail... so I think `throw new RuntimeException` is a correct internal fail.

Can probably use RuntimeException in setQuota as well.

iirc, ceph uses 0 as the value for "no quota"; I don't ecall if that is just a magic value for setting or if it might be returned...  that should never be returned from this method as-is. If there is no quota then the return value must be null.

cavern/src/main/java/org/opencadc/cavern/nodes/CephFSQuotaPlugin.java
```diff
@@ -0,0 +1,158 @@
...
+    }
+
+    @Override
+    public void setQuota(Path directory, Long quota) {
+        LOGGER.debug("setQuota: " + directory);
+        if (directory == null) {
+            throw new IllegalArgumentException("Attempt to set quota on null directory.  Doing nothing.");
+        }
+
+        final String quotaValue = (quota == null || quota == 0L) ? null : Long.toString(quota);
```
pdowler, line 149: w.r.t. quota == 0: I do recall readiung something about using 0 for the ceph attr, but that should be an implementation detail. The only way to remove the quota should be calling this methid with `quota == null`. That's the method API here. 


I don't know if we have a use case to set  a quota to 0 (completely block writes) so I might even throw an IllegalArgumentException if `quota <= 0` for now, just to make it clearly invalid.

pdowler (changes requested):
looks like you need to install the ubuntu attr tools in the CI build. will also need to add them to the cavern Dockerfile (in fedora it is the `attr` package)

while in there, it would be good to update to java 11.

cadc-util-fs/src/main/java/org/opencadc/util/fs/ExtendedFileAttributes.java
```diff
@@ -78,6 +78,8 @@
 
 import java.util.Map;
 import java.util.TreeMap;
+
+import ca.nrc.cadc.util.StringUtil;
```
pdowler, line 82: unused import, plus I know checkstyle doesn't complain much about import style, but we are trying to have ordered imports and no whitespace; can you configure your IDE for that? I'll have to make a task somewhere to improve the checkstyle... otherwise it's formatting churn when different people edit.

I didn't look at import style in  other code (diff doesn't show any issues); only noticed here because there was a new import and no new usage of that class.

pdowler (changes requested):
This looks good. Will aim for a release early next week.

one cosmetic change.

The error handling in the cavern PutAction is hard to evaluate and I don't know off hand how vcp will handle that exception (http 4xx code)... probably need to try it out manually

at88mph:
OK, I will test `vcp` today.  The Checkstyle changes are complete.

at88mph:
I also had an interesting case where I reset my `ceph.quota.max_bytes` to something smaller to test the quota writes, but I neglected to reset the `user` namespace quota.  When the `Node` loaded, it had the `user` namespace quota value in the `NodeProperty`, and wouldn't reset to the `QuotaPlugin`'s value, hence the removal of the old value on line 548 of `NodeUtil`.

pdowler:
setting the plain quota prop in `user` space was a mistake (users could modify it) from previous code; we should probably manually look for and remove that user prop. the workaround is OK for now

pdowler (approved):

PR #246, opened 2024-09-03
enable cavern to handle links to external files

pdowler:
Use file:///path/to/something for an external file (external to cavern root)

allow GET and PUT (create) external file links
reject all attempts to navigate external links (to directories)
reject download of an external file via files or transfer neg.

one optional create + get test added
currently no automated tests of the reject scenarios... TBD


brianmajor (approved):
Pretty straightforward in the end.

PR #248, opened 2024-09-17
cadc-vos-server: fix admin permission check

pdowler:
change to allow admin to update an allocation so quota can be updated


cadc-vos-server/src/main/java/org/opencadc/vospace/server/actions/UpdateNodeAction.java
```diff
@@ -122,7 +122,12 @@ public void doAction() throws Exception {
         }
 
         Subject caller = AuthenticationUtil.getCurrentSubject();
-        if (!voSpaceAuthorizer.hasSingleNodeWritePermission(serverNode, caller)) {
+        if (serverNode instanceof ContainerNode 
+                && nodePersistence.isAllocation((ContainerNode) serverNode) 
+                && Utils.isAdmin(caller, nodePersistence)) {
+            log.debug("write permission granted to admin " + caller);
```
andamian, line 128: I would put a comment as to why we are doing this so we remember later.

andamian (approved):

andamian (approved):

PR #254, opened 2025-06-12
feat: create user allocation as user with jwt

at88mph:
# Description
When allocating a new folder as the root folder admin user, using an OIDC token (JSON Web Token) for a user (creator), ensure the JWT is authenticated and properties are set accordingly.

# Changes
- Minor typos
- Small fix to prevent NPE
- Add new Node property for the creator's JWT (removed before persistence)
- Authentication validation and augmentation

### DIFF
--- a/cadc-vos-server/src/main/java/org/opencadc/vospace/server/actions/CreateNodeAction.java
+++ b/cadc-vos-server/src/main/java/org/opencadc/vospace/server/actions/CreateNodeAction.java
@@ -134,6 +134,9 @@ public class CreateNodeAction extends NodeAction {
         IdentityManager im = AuthenticationUtil.getIdentityManager();
         clientNode.parent = parent;
 
+        // Ensure the caller is set.  It MAY be overridden below.
+        clientNode.owner = caller;
+
         // Extract here to remove later.
         NodeProperty creatorJWT = clientNode.getProperty(VOS.PROPERTY_URI_CREATOR_JWT);
         if (Utils.isAdmin(caller, nodePersistence)) {
@@ -176,8 +179,6 @@ public class CreateNodeAction extends NodeAction {
                 clientNode.ownerID = null; // just in case
 
                 log.debug("Creating user allocation on behalf of OIDC account: OK");
-            } else {
-                clientNode.owner = caller;
             }
         }
 
--- a/cadc-vos-server/src/main/java/org/opencadc/vospace/server/actions/UpdateNodeAction.java
+++ b/cadc-vos-server/src/main/java/org/opencadc/vospace/server/actions/UpdateNodeAction.java
@@ -132,7 +132,6 @@ public class UpdateNodeAction extends NodeAction {
             throw NodeFault.PermissionDenied.getStatus(Utils.getPath(serverNode));
         }
 
-        // TODO: attempt to set owner if admin
         Node storedNode = updateProperties(serverNode, clientNode, nodePersistence, caller);
         IdentityManager im = AuthenticationUtil.getIdentityManager();
         storedNode.ownerDisplay = im.toDisplayString(storedNode.owner);
@@ -149,6 +148,12 @@ public class UpdateNodeAction extends NodeAction {
             throws Exception {
         // merge properties that are Node fields and in Node.properties set
         // TODO: admin could in principle change owner
+
+        // Remove the special Creator JWT property from the client node before copying into the server node as it will never be persisted.
+        NodeProperty creatorJWT = clientNode.getProperty(VOS.PROPERTY_URI_CREATOR_JWT);
+        if (creatorJWT != null) {
+            clientNode.getProperties().remove(creatorJWT);
+        }
         
         if (clientNode.clearReadOnlyGroups || !clientNode.getReadOnlyGroup().isEmpty()) {
             serverNode.getReadOnlyGroup().clear();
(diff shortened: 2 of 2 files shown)


cadc-vos-server/src/main/java/org/opencadc/vospace/server/actions/CreateNodeAction.java
```diff
@@ -132,25 +133,54 @@ public void doAction() throws Exception {
...
-                // network username for toDisplayString(Subject) so that's the value that
-                // normally gets put into node.ownerDisplay and rendered in node documents
-                // so chosing HttpPrincipal here makes output and input (set owner) use the
-                // same values... but seems kind of sketchy.
-                tmp.getPrincipals().add(new HttpPrincipal(clientNode.ownerDisplay));
-                clientNode.owner = im.augment(tmp);
-            } catch (Exception ex) {
-                log.error("failed to map " + clientNode.ownerDisplay + " to a known user", ex);
-                throw ex;
+        if (Utils.isAdmin(caller, nodePersistence)) {
```
pdowler, line 136: With this check on the outside, we need to include the JWT property key to the set of ADMIN_PROPS in the NodePersistence implementation (cavern and vault) so that if anyone else tries to set them they always get filtered out.

I would also do the validation first, eg
```
if (clientNode.ownerDisplay != null && creatorJWT != null) {
    // reject
}
```

rather than down in an `else` block (those get lost very easily). It also keeps the subsequent logic a little simpler.

In general, both the ownerDiplay -> HttpPrincipal and JWT approach are potentially sketchy in their own ways. I would prefer if ther order had the current code first and the new code later, eg:
```
if (Utils.isAdmin(...)) {
    // validate input

    if (clientNode.ownerDisplay != null) {
      // existing code
    } else if (creatorJWT != null) {
      // new code
    } // else: admin creating a container they own: OK
}
```

cadc-vos-server/src/main/java/org/opencadc/vospace/server/actions/CreateNodeAction.java
```diff
@@ -132,25 +133,54 @@ public void doAction() throws Exception {
...
+                    clientNode.owner = nodeOwner;
+                    clientNode.ownerID = null; // just in case
+                } else {
+                    // The "creator" property clashes with the "creator-jwt" property.
+                    throw NodeFault.InvalidArgument.getStatus("BUG: " + VOS.PROPERTY_URI_CREATOR + " property already exists, but " + VOS.PROPERTY_URI_CREATOR_JWT
+                                                                  + " is present. This should not happen.");
+                }
+
+                // remove the creator JWT property if it exists
+                clientNode.getProperties().remove(creatorJWT);
```
pdowler, line 162: note: this still necessary if key added to ADMIN_PROPS

pdowler (changes requested):

cadc-vos-server/src/main/java/org/opencadc/vospace/server/actions/CreateNodeAction.java
at88mph: Added to the `ADMIN_PROPS` as well as the `IMMUTABLE_PROPS` to emulate the `creator` property.

cadc-vos-server/src/main/java/org/opencadc/vospace/server/actions/CreateNodeAction.java
```diff
@@ -132,25 +133,52 @@ public void doAction() throws Exception {
...
+
+                log.debug("Creating user allocation on behalf of OIDC account: OK");
+            } else {
+                clientNode.owner = caller;
             }
-        } else {
-            clientNode.owner = caller;
         }
-        
+
```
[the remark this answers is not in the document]
at88mph: Done.

cadc-vos-server/src/main/java/org/opencadc/vospace/server/actions/CreateNodeAction.java
```diff
@@ -132,25 +133,58 @@ public void doAction() throws Exception {
...
+                nodeOwner = AuthenticationUtil.validateSubject(nodeOwner);
+                nodeOwner = AuthenticationUtil.augmentSubject(nodeOwner);
+
+                log.debug("Setting new allocation node owner to " + im.toDisplayString(nodeOwner));
+                clientNode.owner = nodeOwner;
+                clientNode.ownerID = null; // just in case
+
+                log.debug("Creating user allocation on behalf of OIDC account: OK");
+            } else {
+                clientNode.owner = caller;
```
pdowler, line 180: This `clientNode.owner = caller` has to be the default as it covers normal "user create node"; it is now inside the `isAdmin` block. 

Maybe the best is to always assign it right before the `if (Utils.isAdmin(...) at line 136

cadc-vos-server/src/main/java/org/opencadc/vospace/server/actions/UpdateNodeAction.java
```diff
@@ -179,6 +179,12 @@ public static Node updateProperties(Node serverNode, Node clientNode, NodePersis
                 scn.inheritPermissions = ccn.inheritPermissions;
             }
         }
+
+        // Remove the special Creator JWT property from the client node before copying into the server node as it will never be persisted.
```
pdowler, line 183: there are two comments above with `TODO` related to admin changing ownership. I would remove the first one in doAction and move this block of code up below the one near the start of this method as that will turn into the admin block if we decide to implement it.

pdowler (changes requested):

cadc-vos-server/src/main/java/org/opencadc/vospace/server/actions/CreateNodeAction.java
at88mph: Good idea, thanks.

PR #256, opened 2025-07-18
feat: allow api key for cavern user allocation access

at88mph:
# Description
Allow an `api-key` Authorization header when performing a User Allocation administrative task.  Also ensure a default quota is set on new allocations.

# Changes
- Log grant (client application) that requested use of API Key
- Expose new configuration from the NodePersistence for caller checking
- Assume root node owner for calls to directory creation when authenticated with API key.

### DIFF
--- a/cadc-vos-server/src/main/java/org/opencadc/vospace/server/NodePersistence.java
+++ b/cadc-vos-server/src/main/java/org/opencadc/vospace/server/NodePersistence.java
@@ -69,10 +69,13 @@
 
 package org.opencadc.vospace.server;
 
+import ca.nrc.cadc.auth.AuthenticationUtil;
 import ca.nrc.cadc.io.ResourceIterator;
 import ca.nrc.cadc.net.TransientException;
 import java.net.URI;
+import java.security.Principal;
 import java.util.Set;
+import javax.security.auth.Subject;
 import org.opencadc.vospace.ContainerNode;
 import org.opencadc.vospace.Node;
 import org.opencadc.vospace.NodeNotSupportedException;
@@ -130,6 +133,27 @@ public interface NodePersistence {
      * @return set of immutable property keys
      */
     public Set<URI> getImmutableProps();
+
+    /**
+     * Get the default value of a Node Property.  This will likely be used by create operations when values are missing.
+     * @param propertyKey   The URI of the property key to get the default value for.
+     * @return  String default value for the given property key, or null if no default value is set.
+     */
+    String getDefaultPropertyValue(final URI propertyKey);
+
+    /**
+     * Check if the given caller can administer allocations.  Used by create operations.
+     * @param caller    The caller subject, used to check permissions.
+     * @return  True if the given caller can create new administer new allocations, false otherwise.
+     */
+    boolean isAdmin(Subject caller);
+
+    /**
+     * Get the admin grant for the given caller. This is used to log the specific API Key call.
+     * @param caller    The caller subject, used to pull the token.
+     * @return  String grant, or null if not applicable or not available.
+     */
+    String getAdminGrant(Subject caller);
 
     /**
      * Get the views supported by this node persistence implementation.
--- a/cavern/src/main/java/org/opencadc/cavern/CavernConfig.java
+++ b/cavern/src/main/java/org/opencadc/cavern/CavernConfig.java
@@ -85,6 +85,8 @@ import java.nio.file.Path;
 import java.nio.file.Paths;
 import java.util.ArrayList;
 import java.util.List;
+import java.util.Map;
+import java.util.stream.Collectors;
 import javax.security.auth.Subject;
 import org.apache.log4j.Logger;
 import org.opencadc.cavern.nodes.NoQuotaPlugin;
@@ -93,6 +95,9 @@ import org.opencadc.cavern.nodes.QuotaPlugin;
 public class CavernConfig {
 
     private static final Logger log = Logger.getLogger(CavernConfig.class);
+
+    // The challenge type (after the authorization header and before the token) for API Key authentication.  This is isolated in Cavern for now.
+    public static final String ALLOCATION_API_KEY_HEADER_CHALLENGE_TYPE = "api-key";
 
     public static final String CAVERN_PROPERTIES = "cavern.properties";
     private static final String CAVERN_KEY = CavernConfig.class.getPackage().getName();
@@ -109,10 +114,20 @@ public class CavernConfig {
     
     public static final String ALLOCATION_PARENT = CAVERN_KEY + ".allocationParent";
 
+    // An API Key representing a client that has administrative access to create User Allocations.  More than one is possible.
+    // Format is <applicationClientName>:<apiKeyToken>
+    public static final String ALLOCATION_API_KEY = CAVERN_KEY + ".allocationAPIKey";
+
+    // Default quota in GB for new allocations, if not specified in the allocation request.
+    public static final String DEFAULT_QUOTA_GB = CAVERN_KEY + ".defaultQuotaGB";
+
     public static final String QUOTA_PLUGIN_IMPLEMENTATION = QuotaPlugin.class.getName();
     
     private final URI resourceID;
     private final List<String> allocationParents = new ArrayList<>();
+
+    private final List<String> allocationAPIKeys = new ArrayList<>();
+    private final double defaultQuotaGB;
     
     private final Path root;
     private final Path secrets;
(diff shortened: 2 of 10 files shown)


cadc-vos-server/src/main/java/org/opencadc/vospace/server/NodePersistence.java
```diff
@@ -131,6 +134,27 @@ public interface NodePersistence {
      */
     public Set<URI> getImmutableProps();
 
+    /**
+     * Get the default value of a Node Property.  This will likely be used by create operations when values are missing.
+     * @param propertyKey   The URI of the property key to get the default value for.
+     * @return  String default value for the given property key, or null if no default value is set.
+     */
+    String getDefaultPropertyValue(final URI propertyKey);
```
pdowler, line 142: Seeing how the implementation provides the defaults (config and this method), then the library gets this and carefully injects the default... I think we got this wrong.

The alternative is that the FileSystemNodePersistence.put(...) would evaluate whether default propetries are applicable and add them. That would leave "default quota" a cavern-specific thing with no extra burden on the library or other implementations.

So, I think we remove this from the interface and move that code from library CreateAction into cavern's FileSystemNodePersistence.put(...)

cadc-vos-server/src/main/java/org/opencadc/vospace/server/NodePersistence.java
```diff
@@ -131,6 +134,27 @@ public interface NodePersistence {
...
+
+    /**
+     * Check if the given caller can administer allocations.  Used by create operations.
+     * @param caller    The caller subject, used to check permissions.
+     * @return  True if the given caller can create new administer new allocations, false otherwise.
+     */
+    boolean isAdmin(Subject caller);
```
pdowler, line 149: In principle, an admin can do several things more general than create allocation:
- create a Node owned by someone else
- add or modify admin-only props (quota)
In future, it will probably be the way to modify owner if/when we need that.

So, at a minimum the method name here is more general than what we intend to allow via api keys. I see that UpdateAction also calls this, but api-key callers won';t get that far because UpdateAction doesn't switch users like CreateAction does... right now that's nominally intentional, but that mismatch will look like a bug in a few months. I can also see prepareData wanting to insure a minimum quota in future (it will know how much data it intends to transfer in) and thus wanting to modify an existing quota.

Let's discuss tomorrow in person.

cavern/src/main/java/org/opencadc/cavern/nodes/PosixIdentityManager.java
```diff
@@ -176,15 +199,30 @@ public Object toOwner(Subject subject) {
...
+                        final String apiKeyValue = parts[1].trim();
+
+                        subject.getPrincipals().remove(tokenPrincipal);
+                        subject.getPublicCredentials().add(new AuthorizationToken(challengeType, apiKeyValue, Collections.emptyList()));
+                    }
+                }
+            }
+        }
+
+        // delegate to the configured IdentityManager, assuming it can handle a null Subject
```
pdowler, line 224: I would have done the call to `wrappedIdentityManager.validate(subject)` first because most of the time that would leave the subject with 0 ATPs remaining and you wouldn't have to scan through the others just to skip them.

So I think just move this up to the top, then do the loop looking for api key ATPs.

pdowler (changes requested):

at88mph:
Rework is ready for review.  No changes needed to `cadc-vos-server` anymore.  Note that the grant is logged differently as well.

cadc-vos-server/src/main/java/org/opencadc/vospace/server/NodePersistence.java
at88mph: Done.

cadc-vos-server/src/main/java/org/opencadc/vospace/server/NodePersistence.java
at88mph: Changes submitted.

cavern/src/main/java/org/opencadc/cavern/nodes/FileSystemNodePersistence.java
```diff
@@ -452,6 +465,14 @@ public Node put(Node node) throws NodeNotSupportedException, TransientException
                     throw new UnsupportedOperationException("link to external resource", ex);
                 }
             }
+        } else if ((node instanceof ContainerNode) && isAllocation((ContainerNode) node)
+                && node.getProperty(VOS.PROPERTY_URI_QUOTA) == null) {
+            final NodeProperty defaultNodeProperty = this.defaultProperties.get(this.defaultProperties.indexOf(new NodeProperty(VOS.PROPERTY_URI_QUOTA)));
+            if (defaultNodeProperty == null) {
+                log.warn("No default quota (" + VOS.PROPERTY_URI_QUOTA + ") configured.");
+            } else {
+                node.getProperties().add(new NodeProperty(VOS.PROPERTY_URI_QUOTA, defaultNodeProperty.getValue()));
```
pdowler, line 474: I'd prefer to keep these sanity checks independent, so 
```
if (node instanceof LinkNode) {
    // check target acceptability
}

if (...) {
    // add default quota
}
```

`log.warn`  is probably wrong: no default is a valid config, certainly with the (default) NoOpQuotaPlugin and some people might want that with the Ceph plugin. I might have a `log.debug` but it should mention the config property, not the node property identifier (uri).

cavern/src/main/java/org/opencadc/cavern/CavernInitAction.java
```diff
@@ -103,6 +106,7 @@ public CavernInitAction() {
 
     @Override
     public void doInit() {
+        initIdentityManager();
```
pdowler, line 109: `initIdentityManager` needs to be called after `initNodePersistence` because it needs a value of `jndiNodePersistence`

cavern/src/main/java/org/opencadc/cavern/CavernInitAction.java
```diff
@@ -138,6 +142,25 @@ private void initNodePersistence() {
...
+        // Assuming there isn't more than one Cavern deployed within a JVM.
+        PosixIdentityManager.JNDI_NODE_PERSISTENCE_PROPERTY = this.jndiNodePersistence;
+
+        // To be used by the PosixIdentityManager to wrap the existing IdentityManager.
+        System.setProperty(PosixIdentityManager.WRAPPED_IDENTITY_MANAGER_CLASS_PROPERTY, configuredIdentityManagerClassName);
+
+        // Override the existing IdentityManager.
+        System.setProperty(IdentityManager.class.getName(), PosixIdentityManager.class.getName());
+
+        log.debug("IdentityManager set to: " + PosixIdentityManager.class.getName() + " from " + configuredIdentityManagerClassName);
```
pdowler, line 162: This I would log at info level and make it clear that 
```
log.info(IdentityManager.class.getName() + " = " + PosixIdentityManager.class.getName() 
     + " delegating to " + configuredIdentityManagerClassName);
```
This code only runs once and it's good to always log the results of init from config.

cavern/src/main/java/org/opencadc/cavern/nodes/FileSystemNodePersistence.java
```diff
@@ -144,16 +149,27 @@ public class FileSystemNodePersistence implements NodePersistence {
     private final CavernConfig config;
     private final boolean localGroupsOnly;
 
+    // Set of default properties for nodes.  This will be built from configuration.
+    private final List<NodeProperty> defaultProperties = new ArrayList<>();
```
pdowler, line 153: do we anticipate other default properties? I think it is very unlikely...

I would have a single explicit `NodeProperty DEFAULT_QUOTA` to start and make it a list later.

cavern/README.md
```diff
@@ -57,6 +57,15 @@ org.opencadc.cavern.resourceID = ivo://{authority}/{name}
 # (optional) identify which container nodes are allocations
 org.opencadc.cavern.allocationParent = {top level node}
 
+# (optional) API Keys with administrative privileges.  Multiple keys can be specified, one per line.
+# Example:
```
pdowler, line 61: I prefer this list of config keys to be pretty minimal (no examples) and then explained in the paragraphs below. An example down with the explanation is fine if it helps.

Also, this is missing the `defaultQuotaGB` which should be optional and listed after the quota plugin (for context).

cavern/src/main/java/org/opencadc/cavern/CavernConfig.java
```diff
@@ -169,6 +184,19 @@ public CavernConfig() {
             allocationParents.add(ap);
         }
 
+        adminAPIKeys.addAll(mvp.getProperty(CavernConfig.ADMIN_API_KEY));
+
+        final String configuredDefaultQuotaGB = mvp.getFirstPropertyValue(CavernConfig.DEFAULT_QUOTA_GB);
+        if (configuredDefaultQuotaGB == null) {
+            throw new InvalidConfigException(CavernConfig.DEFAULT_QUOTA_GB + " must be specified");
```
pdowler, line 191: default quota must be optional so the only invalid config is a failure to parse as a double (which you have below).

cavern/src/main/java/org/opencadc/cavern/CavernConfig.java
```diff
@@ -184,6 +212,20 @@ public List<String> getAllocationParents() {
...
+     * @return  Map of allocation API keys, where the key is the application client name, and the value is the API key token.
+     */
+    public Map<String, String> getAdminAPIKeys() {
+        return this.adminAPIKeys.stream().map(apiKey -> apiKey.split(":")).collect(
+            Collectors.toMap(splitKey -> splitKey[0], splitKey -> splitKey[1]));
+    }
+
+    public long getDefaultQuotaBytes() {
+        // Convert to bytes as that's what the NodeProperty expects.  This assumes that the default quota is in GiB.
+        return (long) (defaultQuotaGB * 1024L * 1024L * 1024L);
```
pdowler, line 226: No need for a method, actually awkward/misleading since the value is optional. Just use a `public final Long defaultQuota` field (can be null) and assign the computed value directly above.

pdowler (changes requested):

pdowler (approved):

PR #273, opened 2026-05-05
Cavern papi support alt

at88mph:
# Changes
- alternative implementation to only call the configured permissions api on user allocation request

### DIFF
--- a/cavern/src/main/java/org/opencadc/cavern/PermissionsClientConfig.java
+++ b/cavern/src/main/java/org/opencadc/cavern/PermissionsClientConfig.java
@@ -1,0 +1,93 @@
+package org.opencadc.cavern;
+
+import ca.nrc.cadc.util.MultiValuedProperties;
+import ca.nrc.cadc.util.StringUtil;
+
+import java.net.MalformedURLException;
+import java.net.URI;
+import java.net.URL;
+
+public class PermissionsClientConfig {
+    private static final String PREFIX = ".permissions";
+    private static final String PERMISSIONS_API_BASE_URL = PREFIX + ".baseUrl";
+    private static final String PERMISSIONS_API_AUTH_BASE_URL = PREFIX + ".authBaseUrl";
+    private static final String PERMISSIONS_API_SERVICE_NAME = PREFIX + ".serviceName";
+    private static final String PERMISSIONS_API_AUTHORISE_TYPE = PREFIX + ".authoriseType";
+    private static final String PERMISSIONS_API_ROUTE = PREFIX + ".route";
+    private static final String PERMISSIONS_API_VERSION = PREFIX + ".version";
+    private static final String PERMISSIONS_API_METHOD = PREFIX + ".method";
+
+    private final String configPrefix;
+
+    private final String permissionsApiBaseUrl;
+    private final String permissionsApiAuthBaseUrl;
+    private final String serviceName;
+    private final String authoriseType;
+    private final String routePath;
+    private final String version;
+    private final String method;
+
+    /**
+     * Read in the configuration from properties.  This will be NULL if no base URL is set (presumed to not be
+     * configured for this Cavern instance).
+     * @param properties    The MultiValuedProperties instance.
+     * @param configPrefix  The cavern prefix to look in the config file.
+     * @return  PermissionsClientConfig instance, or null if not configured.
+     */
+    public static PermissionsClientConfig fromProperties(final MultiValuedProperties properties, final String configPrefix) {
+        final String baseUrl = properties.getFirstPropertyValue(configPrefix + PERMISSIONS_API_BASE_URL);
+        return StringUtil.hasText(baseUrl)
+                ? new PermissionsClientConfig(properties, configPrefix)
+                : null;
+    }
+
+    private PermissionsClientConfig(final MultiValuedProperties properties, final String configPrefix) {
+        this.configPrefix = configPrefix;
+
+        this.permissionsApiBaseUrl = properties.getFirstPropertyValue(configPrefix + PERMISSIONS_API_BASE_URL);
+        this.permissionsApiAuthBaseUrl = properties.getFirstPropertyValue(configPrefix + PERMISSIONS_API_AUTH_BASE_URL);
+        this.serviceName = properties.getFirstPropertyValue(configPrefix + PERMISSIONS_API_SERVICE_NAME);
+        this.authoriseType = properties.getFirstPropertyValue(configPrefix + PERMISSIONS_API_AUTHORISE_TYPE);
+        this.routePath = properties.getFirstPropertyValue(configPrefix + PERMISSIONS_API_ROUTE);
+        this.version = properties.getFirstPropertyValue(configPrefix + PERMISSIONS_API_VERSION);
+        this.method = properties.getFirstPropertyValue(configPrefix + PERMISSIONS_API_METHOD);
+    }
+
+    public URL getPermissionsApiBaseUrl() {
+        try {
+            return URI.create(this.permissionsApiBaseUrl).toURL();
+        } catch (MalformedURLException malformedURLException) {
+            throw new IllegalStateException("Invalid URL for " + this.configPrefix + PERMISSIONS_API_BASE_URL + ": "
+                    + this.permissionsApiBaseUrl, malformedURLException);
+        }
+    }
+
+    public URL getPermissionsApiAuthBaseUrl() {
+        try {
+            return URI.create(this.permissionsApiAuthBaseUrl).toURL();
+        } catch (MalformedURLException malformedURLException) {
+            throw new IllegalStateException("Invalid URL for " + this.configPrefix + PERMISSIONS_API_AUTH_BASE_URL + ": "
+                    + this.permissionsApiAuthBaseUrl, malformedURLException);
+        }
+    }
+
+    public String getServiceName() {
+        return serviceName;
+    }
+
+    public String getAuthoriseType() {
+        return authoriseType;
+    }
+
+    public String getRoutePath() {
+        return routePath;
+    }
+
+    public String getVersion() {
+        return version;
+    }
+
+    public String getMethod() {
+        return method;
+    }
+}
--- a/cavern/src/main/java/org/opencadc/cavern/PermissionsClientConfig.java
+++ b/cavern/src/main/java/org/opencadc/cavern/PermissionsClientConfig.java
@@ -7,24 +7,22 @@ import java.net.URI;
 import java.net.URL;
 
 public class PermissionsClientConfig {
-    private static final String PREFIX = ".permissions";
+    private static final String PREFIX = ".papi";
     private static final String PERMISSIONS_API_BASE_URL = PREFIX + ".baseUrl";
     private static final String PERMISSIONS_API_AUTH_BASE_URL = PREFIX + ".authBaseUrl";
-    private static final String PERMISSIONS_API_SERVICE_NAME = PREFIX + ".serviceName";
-    private static final String PERMISSIONS_API_AUTHORISE_TYPE = PREFIX + ".authoriseType";
-    private static final String PERMISSIONS_API_ROUTE = PREFIX + ".route";
-    private static final String PERMISSIONS_API_VERSION = PREFIX + ".version";
-    private static final String PERMISSIONS_API_METHOD = PREFIX + ".method";
 
     private final String configPrefix;
 
     private final String permissionsApiBaseUrl;
     private final String permissionsApiAuthBaseUrl;
-    private final String serviceName;
-    private final String authoriseType;
-    private final String routePath;
-    private final String version;
-    private final String method;
+
+    // Hard-coded values for now.
+    private final String serviceName = "canfar-api";
+    private final String authoriseType = "route";
+    private final String routePath = "/nodes/home/{username}";
+    private final String version = "1";
+    private final String method = "PUT";
+
 
     /**
      * Read in the configuration from properties.  This will be NULL if no base URL is set (presumed to not be
@@ -45,11 +43,6 @@ public class PermissionsClientConfig {
 
         this.permissionsApiBaseUrl = properties.getFirstPropertyValue(configPrefix + PERMISSIONS_API_BASE_URL);
         this.permissionsApiAuthBaseUrl = properties.getFirstPropertyValue(configPrefix + PERMISSIONS_API_AUTH_BASE_URL);
-        this.serviceName = properties.getFirstPropertyValue(configPrefix + PERMISSIONS_API_SERVICE_NAME);
-        this.authoriseType = properties.getFirstPropertyValue(configPrefix + PERMISSIONS_API_AUTHORISE_TYPE);
-        this.routePath = properties.getFirstPropertyValue(configPrefix + PERMISSIONS_API_ROUTE);
-        this.version = properties.getFirstPropertyValue(configPrefix + PERMISSIONS_API_VERSION);
-        this.method = properties.getFirstPropertyValue(configPrefix + PERMISSIONS_API_METHOD);
     }
 
     public URL getPermissionsApiBaseUrl() {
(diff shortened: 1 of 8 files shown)


cavern/src/main/java/org/opencadc/cavern/actions/CreateNodeAction.java
```diff
@@ -0,0 +1,100 @@
...
+import org.opencadc.vospace.ContainerNode;
+import org.opencadc.vospace.Node;
+
+/**
+ * Overridden create action to allow the use of a permissions client to access a remote service and do verification
+ * that the caller has permissions to create an allocation.  Administrative access is then assumed.
+ */
+public class CreateNodeAction extends org.opencadc.vospace.server.actions.CreateNodeAction {
+    @Override
+    public void doAction() throws Exception {
```
pdowler, line 24: Given the _elevate privilege_ nature of this, I would like to see this organised for easier review and future code changes. Maybe:
```
doAction() ...{
    Node targetNode = getInputNode();
    Subject caller = AuthenticationUtil.getCurrentSubject();
    if (isSelfAllocation(caller, targetNode) {
         checkSelfAllocatePermission(caller, targetNode);
         Subject.doAs(fileSystemNodePersistence.getRootNode().owner, ...
         return;     
    }
    // normal create
    super.doAction();
}
```
so we can separate checking if the call is a self-allocation (should be just local code), the permissions check, and the elevated call to the base class. That keeps the dependence on external services (PAPI or our own in that one permission check method, which either throws or returns (to proceed).

cavern/src/main/java/org/opencadc/cavern/PermissionsClientConfig.java
```diff
@@ -0,0 +1,92 @@
+package org.opencadc.cavern;
+
+import ca.nrc.cadc.util.MultiValuedProperties;
+import ca.nrc.cadc.util.StringUtil;
+import java.net.MalformedURLException;
+import java.net.URI;
+import java.net.URL;
+
+public class PermissionsClientConfig {
+    private static final String PREFIX = ".permissions";
```
pdowler: First thing: config has to be documented in the README, but...

This is a lot of extra config for `cavern.properties` and will substantially make the cavern README worse without adding much value

Are the two base URLs for PAPI and AuthAPI (token exchange) respectively?

**Is it feasible to just configure the base URL(s) and hard-code the rest?** Can different deployments actually chose different values, or will they all be the same?

At some point in future we may want to hook into CADC AAI services, so maybe this config key should be more specific, eg
```
org.opencadc.cavern.papiBaseURL = {PAPI}
```
reason: a single more generic config would be viable if the service had a VOSI-capabilities endpoint and we could figure out which API it implements - then we could read caps and create the approp client.

pdowler (changes requested):
initial review

cavern/src/main/java/org/opencadc/cavern/PermissionsClientConfig.java
at88mph: Here is a sample from the Helm Chart's configuration:
```yaml
                    permissionsAPI:
                      enabled: true
                      baseURL: "https://permissions.srcnet.skao.int/api"
                      authAPIBaseURL: "https://authn.srcnet.skao.int/api"
                      type: "route"
                      route: "/cavern/nodes/home/{username}"
                      name: "canfar-api"
                      version: "1"
                      method: "POST"
```
The baseURL for PAPI and the auth URL are different in the current production setup.  Do they belong in the registry?  Unclear.

The rest _could_ be hardcoded, especially the name, version, type, and method, I guess.

cavern/src/main/java/org/opencadc/cavern/actions/CreateNodeAction.java
at88mph: Sure.  Is this approach preferred then?  I provided two as a matter of interest, but if you have alternates please let me know.

pdowler: I will look at the other PR now...

pdowler: So PR #272 relies on the internal PosixIdentityManager to elevate caller privilege if PermissionsClient is configured? And then return the admin user if the route is authorised? 

I don't think this is viable:
- PosixIdentityManager doesn't have the context (request is an attempt to self-allocate) to decide it should elevate
- I think it will always call PAPI and rely on PAPI to reject most create-node requests: risky and adds failure modes

So yes, I think doing this in CreateNodeAction where sufficient context is available is the better approach.

cavern/src/main/java/org/opencadc/cavern/PermissionsClientConfig.java
pdowler: For the example `route` this asserts that a user can POST to `/cavern/nodes/home/{username}`?
- cavern has to fill in the {username}?
- how will PAPI handle the fact that any deployment can rename `cavern`, eg to `arc` or something else?
- why is it POST when create node is a PUT?

One possibility of "something else" is to war rename cavern.war to ROOT.war and use a distinct hostname, which is the style that papi and authn use.... then the `route` might be `/nodes/home/{username}`

cavern/src/main/java/org/opencadc/cavern/actions/CreateNodeAction.java
at88mph: That class is cleaned up now.

cavern/src/main/java/org/opencadc/cavern/PermissionsClientConfig.java
at88mph: OK, the `/nodes/home/{username}` might be a better fit then.  The `POST` should be a `PUT`, you're right.

I'm pretty sure that the `{username}` does need to be filled in, and that's just a wildcard for PAPI.  I'm waiting to hear from Rob Barnsley and/or James Collinson.

at88mph: In retrospect though, that path is just a string passed into configuration.  If they configure they're WAR rename to be /my-cavern, the configuration can still be `/cavern/nodes/home/{username}`.

cavern/src/main/java/org/opencadc/cavern/actions/CreateNodeAction.java
```diff
@@ -0,0 +1,108 @@
...
+            final FileSystemNodePersistence fileSystemNodePersistence = getFileSystemNodePersistence();
+            final Subject validatedSubject = validateCurrentSubject(fileSystemNodePersistence);
+            final CavernConfig config = fileSystemNodePersistence.getConfig();
+            final PermissionsClientConfig permissionsClientConfig = config.getPermissionsClientConfig();
+            final Node inputNode = getInputNode();
+
+            if (shouldUsePermissionsClient(permissionsClientConfig, inputNode, fileSystemNodePersistence)) {
```
pdowler: I really want isSelfAllocation to be followed by the simple but explicit permission check and then a Subject.doAs(...) here so the elevated privilege is clear. That makes the sequence as clear as possible and means the general concept is not written as an srcnet/papi-specific hack but a correctly thought-out feature.

All code related accessing the permissions config should be hidden inside the permission check method. That way we can have that code do different checks depending on how cavern is configured without it showing up in the logic of doAction. So no PermissionsConfig, AuthorisationResult, etc in this method... just a method that performs the authorisation check and throws the AccessControlException.
aside: the current subject is already validated so you do not need to do that again.

**so:** please move permission  check into a method that might throw the AccessControlException
**and:** explicit Subjectg.doAs(...) right in doAction()

I anticipate that if you implement isSelfAllocation it will become clear that we do not have enough information to completely do that: we do not know what the users home  dir should be and we cannot differentiate between a home dir and a project dir.  Once this is clear, we can figure out how to deal with that specific issue.

future: I hope to refactor most of self-allocate up to the library and not need a custom CreateNodeAction (because overriding doAction() is flexible but can be brittle).

pdowler (changes requested):

cavern/src/main/java/org/opencadc/cavern/actions/CreateNodeAction.java
at88mph: Yes, this got reworked immediately after this commit.  Can you check again, please?

cavern/src/main/java/org/opencadc/cavern/actions/CreateNodeAction.java
```diff
@@ -0,0 +1,113 @@
...
+    private FileSystemNodePersistence getFileSystemNodePersistence() {
+        // Should never happen, but here for completeness.
+        if (this.nodePersistence instanceof FileSystemNodePersistence) {
+            return (FileSystemNodePersistence) this.nodePersistence;
+        }
+
+        throw new IllegalStateException("Node persistence is not an instance of FileSystemNodePersistence");
+    }
+
+    private Subject validateCurrentSubject(final FileSystemNodePersistence fileSystemNodePersistence)
```
pdowler: validateCurrentSubject is unnecessary

cavern/src/main/java/org/opencadc/cavern/actions/CreateNodeAction.java
```diff
@@ -0,0 +1,113 @@
...
+            throws InvalidConfigException {
+        return fileSystemNodePersistence.getIdentityManager().validate(AuthenticationUtil.getCurrentSubject());
+    }
+
+    private boolean isSelfAllocation(final Subject callingSubject,
+                                     final Node inputNode,
+                                     final FileSystemNodePersistence fileSystemNodePersistence) {
```
pdowler: This method can use the `nodePersistence` API directly, so can be removed from method args and used directly.

cavern/src/main/java/org/opencadc/cavern/actions/CreateNodeAction.java
```diff
@@ -0,0 +1,113 @@
...
+ */
+public class CreateNodeAction extends org.opencadc.vospace.server.actions.CreateNodeAction {
+    @Override
+    public void doAction() throws Exception {
+        final FileSystemNodePersistence fileSystemNodePersistence = getFileSystemNodePersistence();
+        final Subject validatedSubject = validateCurrentSubject(fileSystemNodePersistence);
+        final Node inputNode = getInputNode();
+
+        if (isSelfAllocation(validatedSubject, inputNode, fileSystemNodePersistence)) {
+            log.debug("Using permissions client to allocate user.");
```
pdowler, line 26: please move the authorization check to a method; this is the only bit that is srcnet/papi specific and I want to isolate it... by returning the AuthorisationResult and seeing it here, all the logic gets tangled with (what should/will be) plugin code. We simply want a method that returns (if OK) or throws AccessControlException.

and so that privilege elevation is clear: replace handleAuthenticatedAction with the direct Subject.doAs(...) call

cavern/src/main/java/org/opencadc/cavern/actions/CreateNodeAction.java
```diff
@@ -0,0 +1,113 @@
...
+
+/**
+ * Overridden create action to allow the use of a permissions client to access a remote service and do verification
+ * that the caller has permissions to create an allocation.  Administrative access is then assumed.
+ */
+public class CreateNodeAction extends org.opencadc.vospace.server.actions.CreateNodeAction {
+    @Override
+    public void doAction() throws Exception {
+        final FileSystemNodePersistence fileSystemNodePersistence = getFileSystemNodePersistence();
+        final Subject validatedSubject = validateCurrentSubject(fileSystemNodePersistence);
```
pdowler: the above is confusing; just call AuthenticationUtil.getCurrentSubject() directly

cavern/src/main/java/org/opencadc/cavern/actions/CreateNodeAction.java
```diff
@@ -0,0 +1,113 @@
...
+    private boolean isSelfAllocation(final Subject callingSubject,
+                                     final Node inputNode,
+                                     final FileSystemNodePersistence fileSystemNodePersistence) {
+        final Subject owner = fileSystemNodePersistence.getRootNode().owner;
+        return !owner.equals(callingSubject)
+                && inputNode instanceof ContainerNode
+                && fileSystemNodePersistence.isAllocation(((ContainerNode) inputNode));
+    }
+
+    private AuthorisationResult authoriseAllocation(final Subject validatedSubject,
```
pdowler: this method should be `void authoriseAllocation(...)` and should check the result and possibly throw AccessControlException

prefer that it does not take the PAPI-specific PermissionsClientConfig as an arg and finds the config itself; then we can feasibly introduce a different authorise logic for our AAI system

pdowler (changes requested):

at88mph:
Thank you.  I've submitted another round of rework.

at88mph:
OK, another rework push.

cavern/src/main/java/org/opencadc/cavern/actions/CreateNodeAction.java
```diff
@@ -0,0 +1,108 @@
...
+            return (FileSystemNodePersistence) this.nodePersistence;
+        }
+
+        throw new IllegalStateException("Node persistence is not an instance of FileSystemNodePersistence");
+    }
+
+    private boolean isSelfAllocation() {
+        final Subject caller = AuthenticationUtil.getCurrentSubject();
+        final Node inputNode = getInputNode();
+        final FileSystemNodePersistence fileSystemNodePersistence = getFileSystemNodePersistence();
```
pdowler, line 68: Subject.equals(Subject) won't work for this... but anyway this can just use `Utils.isAdmin(caller, nodePersistence)` (from cadc-vos-server)

Also, can just use `nodePersistence.isAllocation(...)` directly without cast.

still thinking about a hack to limit this to a home dir...

cavern/src/main/java/org/opencadc/cavern/PermissionsClientConfig.java
```diff
@@ -0,0 +1,85 @@
...
+
+    private final String configPrefix;
+
+    private final String permissionsApiBaseUrl;
+    private final String permissionsApiAuthBaseUrl;
+
+    // Hard-coded values for now.
+    private final String serviceName = "canfar-api";
+    private final String authoriseType = "route";
+    private final String routePath = "/nodes/home/{username}";
```
pdowler: remove `/nodes` from routePath

cavern/src/main/java/org/opencadc/cavern/actions/CreateNodeAction.java
```diff
@@ -0,0 +1,108 @@
...
+        final Subject owner = fileSystemNodePersistence.getRootNode().owner;
+        return !owner.equals(caller)
+                && inputNode instanceof ContainerNode
+                && fileSystemNodePersistence.isAllocation(((ContainerNode) inputNode));
+    }
+
+    private void authorize() throws Exception {
```
pdowler: change to `private boolean authorize()` and return true only if PAPI authorises.

fall through: false

cavern/src/main/java/org/opencadc/cavern/actions/CreateNodeAction.java
```diff
@@ -0,0 +1,108 @@
...
+import org.opencadc.vospace.Node;
+
+/**
+ * Overridden create action to allow the use of a permissions client to access a remote service and do verification
+ * that the caller has permissions to create an allocation.  Administrative access is then assumed.
+ */
+public class CreateNodeAction extends org.opencadc.vospace.server.actions.CreateNodeAction {
+    @Override
+    public void doAction() throws Exception {
+        if (isSelfAllocation()) {
```
pdowler: change to `if (isSelfAllocation() && authorize())` so this falls through to normal super

use early return after Subject.doAs so we can get rid of else (lower complexity style, especially applicable to special behaviour like this)

pdowler (changes requested):

cavern/src/main/java/org/opencadc/cavern/actions/CreateNodeAction.java
at88mph: The `isAllocation` method does expect a `ContainerNode`, so a cast is necessary.

at88mph:
Rework added.

at88mph:
Replaced by #274

PR #274, opened 2026-05-25
Cavern papi support alt - refactored into cadc-vos-server

pdowler:
I have not documented the configuration but by configuring a `selfAllocateGroup` I was able to make all the rejections happen as intended.

I still have to fix the cadc-test-vos setup because it relied on `allocationParent=/`.

### DIFF
--- a/cadc-vos-server/src/main/java/org/opencadc/vospace/server/actions/CreateNodeAction.java
+++ b/cadc-vos-server/src/main/java/org/opencadc/vospace/server/actions/CreateNodeAction.java
@@ -174,6 +174,7 @@ public class CreateNodeAction extends NodeAction {
         // backwards compat: ignore if clients still send this, just in case
         NodeProperty creatorJWT = clientNode.getProperty(VOS.PROPERTY_URI_CREATOR_JWT);
         if (creatorJWT != null) {
+            log.warn("found obsolete " + VOS.PROPERTY_URI_CREATOR_JWT + " -- DROPPED");
             clientNode.getProperties().remove(creatorJWT);
         }
 
--- a/cadc-vos-server/src/main/java/org/opencadc/vospace/server/actions/CreateNodeAction.java
+++ b/cadc-vos-server/src/main/java/org/opencadc/vospace/server/actions/CreateNodeAction.java
@@ -3,7 +3,7 @@
  *******************  CANADIAN ASTRONOMY DATA CENTRE  *******************
  **************  CENTRE CANADIEN DE DONNÉES ASTRONOMIQUES  **************
  *
- *  (c) 2023.                            (c) 2023.
+ *  (c) 2026.                            (c) 2026.
  *  Government of Canada                 Gouvernement du Canada
  *  National Research Council            Conseil national de recherches
  *  Ottawa, Canada, K1A 0R6              Ottawa, Canada, K1A 0R6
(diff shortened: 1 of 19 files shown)


cadc-vos-server/src/main/java/org/opencadc/vospace/server/actions/CreateNodeAction.java
```diff
@@ -124,66 +127,52 @@ public void doAction() throws Exception {
...
-        // Ensure it's always removed after being handled in the admin block.  This property is never persisted, as it's just a means to an end for the
-        // creator property.
+        if (!voSpaceAuthorizer.hasSingleNodeWritePermission(parent, writePerm)) {
+            throw NodeFault.PermissionDenied.getStatus(clientNodeURI.toString());
+        }
+        // protected against accidentally using this again
+        writePerm = null;
+        
+        // backwards compat: ignore if clients still send this, just in case
+        NodeProperty creatorJWT = clientNode.getProperty(VOS.PROPERTY_URI_CREATOR_JWT);
```
at88mph, line 175: This is probably worth logging, too.  I can see previous incarnations wondering why it's not working anymore.

cavern/src/main/java/org/opencadc/cavern/actions/CreateNodeAction.java
```diff
@@ -0,0 +1,113 @@
...
+        final CavernConfig config = fileSystemNodePersistence.getConfig();
+        
+        // two mechanisms to authorize self-allocate:
+        // cavern condfigured with self-allocate group
+        // cavern configured to use SRCNet Permissions API
+        final Set<GroupURI> allowedGroups = config.getSelfAllocateGroups();
+        final PermissionsClientConfig permissionsClientConfig = config.getPermissionsClientConfig();
+        if (allowedGroups.isEmpty() && permissionsClientConfig == null) {
+            log.debug("self-allocate mechanism not configured -- could be a misconfiguration or intentional");
+            super.checkSelfAllocatePermission(caller, uri);
```
at88mph, line 48: The super method just throws an `AccessControlException`, but it looks like the intent is to pass it through as potentially being intentional.  I actually like the idea of Cavern being opinionated about self-allocation, and simply failing here, but maybe with an `InvalidConfigException` to point out that it's needed.

at88mph (changes requested):

cadc-vos-server/src/main/java/org/opencadc/vospace/server/actions/CreateNodeAction.java
pdowler: added `log.warn`

cavern/src/main/java/org/opencadc/cavern/actions/CreateNodeAction.java
pdowler: Invalid config is up to CavernConfig to decide - that would be attempting to configure self-allocate auth but doing it wrong.

Here, it is simple not configured, which is the default/normal case for anyone deploying cavern... calling super method defers the normal case to the library, which is the safe case. The only thing that could be changed is the message to the user (by throwing AccessControlException directly) and maybe trying to finesse the message in the log event. 

I'm going to see if the log messages are useful enough, but I think the AccessControlException will replace the earlier `self-allocate` message, not add to it.

at88mph (approved):

at88mph:
Tested in a test Kubernetes Cluster and works as expected.

PR #276, opened 2026-05-28
move PAPI config to LocalAuthority

pdowler:


at88mph (approved):
That's as expected.  LGTM

PR #277, opened 2026-06-09
Recursive Runner class to get report on dir size

aratikakadiya:


cadc-vos-server/src/main/java/org/opencadc/vospace/server/async/RecursiveNodeSizeRunner.java
```diff
@@ -0,0 +1,289 @@
...
+
+        // TODO: Need to check if the bytesUsed is in sync. If not, need to remove this block of code to keep the recursion going.
+        if (depth >= maxDepth && cn.bytesUsed != null) {
+            incSuccessCount();
+            addToReport(Utils.getPath(cn), cn.bytesUsed, depth);
+            return cn.bytesUsed;
+        }
+
+        long currentNodeSize = 0L;
+        try (ResourceIterator<Node> iter = nodePersistence.iterator(cn, null, null)) {
```
pdowler: Look closely at the RecursiveNodePropsRunner.performAction - when it iterates over a ContainerNode, it simply collects the child CNs in a list and processes the DataNode(s). Then once that iterator is done, it performs the recursion (line 170). This ensures that only one iterator is _open_ at one time *aka* in vault there is one database query active at a time. 

Without this, either we re-use the same db connection and flush a lot of results to the client side *or* we use multiple connections and that leads to a dead-lock when we run out of connections in the connection pool... so it's imperative that we only use one db connection at a time.

cadc-vos-server/src/main/java/org/opencadc/vospace/server/async/NodeSizeReportServlet.java
```diff
@@ -0,0 +1,234 @@
+/*
```
pdowler: It looks like you realised this later (in vault): there is no need for a new Servlet class: the new JobRunner is deployed using a standard servlet from `cadc-rest` (well, actually the JobServlet) and action classes from `cadc-uws-server`. Please remove this class.

cadc-vos-server/src/main/java/org/opencadc/vospace/server/async/RecursiveNodeSizeRunner.java
```diff
@@ -0,0 +1,289 @@
...
+            while (iter.hasNext()) {
+                Node child = iter.next();
+                if (child instanceof ContainerNode) {
+                    Long childNodeSize = accumulateNodeSize(child, subject, depth + 1);
+                    if (childNodeSize > 0L) {
+                        currentNodeSize += childNodeSize;
+                    }
+                } else {
+                    if (!authorizer.hasSingleNodeReadPermission(child, subject)) {
```
pdowler: I think this permission check is not needed: if the caller has read permission on the parent container they are allowed to see the existence and sizes of child nodes. 

Also, there are also LinkNode(s) so the `else` above should check that this is actually a DataNode. A comment at the end like `// else: ignore LinkNode` would probably be a good addition as well.

cadc-vos-server/src/main/java/org/opencadc/vospace/server/async/RecursiveNodeSizeRunner.java
```diff
@@ -0,0 +1,289 @@
...
+        if (Thread.currentThread().isInterrupted()) {
+            throw new InterruptedException();
+        }
+
+        checkJobPhase();
+
+        if (!authorizer.hasSingleNodeReadPermission(node, subject)) {
+            log.debug("no read permission on node " + Utils.getPath(node));
+            // incErrorCount(); // Not considering this as an error. Just reporting permission denied for this node and continuing with the rest of the tree.
+            addToReport(Utils.getPath(node), -1L, depth);
```
pdowler: I think this will not report the permission denied when farther down in the tree than max depth, but that's probably necessary to be visible to the caller.

cadc-vos-server/src/main/java/org/opencadc/vospace/server/async/RecursiveNodeSizeRunner.java
```diff
@@ -0,0 +1,289 @@
...
+        return nodeSize >= 0L; // success if permissions are OK
+    }
+
+    private void addToReport(String path, long size, int depth) {
+        if (depth <= maxDepth) {
+            reportLines.put(path, size);
+        }
+    }
+
+    private Long accumulateNodeSize(Node node, Subject subject, int depth) throws Exception {
```
pdowler: If you changed the method signature to
```
long accumulateNodeSize(ContainerNode node, ...)
```
the code below would be simpler and more clear. You already have a suitable check at the start of performAction.

And I think it could be primitive `long` but didn't look too closely at whether that's more or less complicated (because auto-unboxing is sometimes a pain).

pdowler (changes requested):
Just review of main logic so far

cadc-vos-server/src/main/java/org/opencadc/vospace/server/async/RecursiveNodeSizeRunner.java
```diff
@@ -0,0 +1,278 @@
...
+        }
+
+        long nodeSize = accumulateNodeSize(root, subject, 0);
+        totalBytes = Math.max(nodeSize, 0);
+        return nodeSize >= 0L; // success if permissions are OK
+    }
+
+    // Note: Report permission denied for all the depths
+    private void addToReport(String path, long size, int depth) {
+        if (depth <= maxDepth || size < 0L) {
```
pdowler: I grok that size == -1 means is from permission denied at line 180-183. I'll have to see what the report looks like in real life to comment further. 

I don't think `accumulateNodeSize` needs to return -1 (line 184) though: it could just return 0 and then down on line 220 you would not actually need to check. Is there something else that -1 does that I am missing?

cadc-vos-server/src/main/java/org/opencadc/vospace/server/async/RecursiveNodeSizeRunner.java
```diff
@@ -0,0 +1,278 @@
...
+import org.opencadc.vospace.VOSURI;
+import org.opencadc.vospace.server.Utils;
+
+/**
+ * Async job runner that computes recursive byte usage for an allocation.
+ */
+public class RecursiveNodeSizeRunner extends AbstractRecursiveRunner {
+
+    private static final Logger log = Logger.getLogger(RecursiveNodeSizeRunner.class);
+
```
pdowler, line 108: It would be good for the input parameters to be declared as constants here for clarity. Lower case values for the parameter names is preferred.

I think we decided that usage of this tool requires a "destination" parameter; in another context we use `dest={uri}` to have async job send output to VOSpace so that would be the preferred name. If the `dest` is in the same VOSpace service this job is running in, the code should use the NodePersistence API to generate a URL to write the result to. 

out of scope: `dest` is external (a different VOSpace or an http URL) could be done but is probably not needed, so `throw new UnsupportedOperationException`...

cadc-vos-server/src/main/java/org/opencadc/vospace/server/async/RecursiveNodeSizeRunner.java
```diff
@@ -0,0 +1,278 @@
...
+    @Override
+    protected boolean performAction(Node node) throws Exception {
+        if (!(node instanceof ContainerNode)) {
+            sendError("target must be a ContainerNode");
+            return false;
+        }
+        ContainerNode root = (ContainerNode) node;
+        if (!nodePersistence.isAllocation(root)) {
+            sendError("target is not an allocation root");
```
pdowler, line 156: I don't think limiting this job to an allocation root container node actually makes this simpler and it does make it less useful. Do you agree with that? Maybe comment out for now to see if that's true.

pdowler (changes requested):

cadc-vos-server/src/main/java/org/opencadc/vospace/server/async/RecursiveNodeSizeRunner.java
aratikakadiya: I agree. 

Jira description : `we will only support doing this on an allocation and only calculating the size of the whole allocation.` So I thought it is only for user allocation. But it sure limits the use of this api so I have commented the code for now. And if we never want to limit it to user allocation, I should remove it asap. lmk.

cadc-vos-server/src/main/java/org/opencadc/vospace/server/async/RecursiveNodeSizeRunner.java
aratikakadiya: Updated. Thanks for pointing out the use of NodePersistance API. I was struggling as I took the long path as the HttpStorageManager not knowing this situation. Saved!

cadc-vos-server/src/main/java/org/opencadc/vospace/server/async/RecursiveNodeSizeRunner.java
aratikakadiya: Nice catch! It was relevant before but after optimization and code changes, it is not relevant. Simplified even more! Thanks.

While writing the report, I am updating it to `Permission Denied`. I need `long` for sorting so I have set -1.

cadc-vos-server/src/main/java/org/opencadc/vospace/server/async/AbstractRecursiveRunner.java
```diff
@@ -308,4 +313,12 @@ protected void sendError(String message) {
                     + " to " + ExecutionPhase.ERROR, t);
         }
     }
+
+    /**
+     * Subclasses may override to attach additional UWS Result entries
+     * (e.g., computed metrics, reports). Default: no extra results.
+     */
+    protected List<Result> getAdditionalResults() {
```
pdowler: Is this needed and used now output is written to a storage URL?

cadc-vos-server/src/main/java/org/opencadc/vospace/server/async/RecursiveNodeSizeRunner.java
```diff
@@ -0,0 +1,336 @@
...
+
+    private long totalBytes = 0;
+    private int maxDepth = 0;
+    private VOSURI dest;
+
+    private Writer out;
+    private Exception scanException = null;
+
+    @Override
+    public void setJob(Job job) {
```
pdowler: It is best to do this at the start of performAction and not in setJob; this method should probably not throw

I realise the other impls do some work in setJob and I will fix that and not override this method.

reason: exception throwing and handling needs to be in JobRunner.run(); the base class does all that and calls performAction() and handles all failures from it

cadc-vos-server/src/main/java/org/opencadc/vospace/server/async/RecursiveNodeSizeRunner.java
```diff
@@ -0,0 +1,336 @@
...
+                    out.flush();
+                } catch (Exception e) {
+                    scanException = e; // save exception for later
+                    throw new RuntimeException(e);
+                }
+            },
+            putURL
+        );
+        upload.setRequestProperty("Content-Type", CONTENT_TYPE);
+        upload.run();
```
pdowler, line 177: I think I grok how the above expression will work. `upload.run()` will make the call to `putURL`, check errors, and then open the output and call OutputStreamWrapper.write(OutputStream). The thing I don't like about the above expression style is it never actually says it is implementing that `write` method and the `stream ->` in the HttpUpload ctor call is very subtle. aside: the `out` variable here is actually a member of the outer class; I usually try to make assignment to class variables stand out with `this.out = ...`.

This version has made the complication of trying to do a lazy init and use OutputStreamWrapper clear. I think there is a way to do it that would work in the typical case (depth=0 and large allocation)...

I think this can be made more clear with a concrete nested class eg:
```
private class DynamicReportOutput implements OutputStreamWrapper {
    
    private PrintWriter output;

    DynamicReportOutput(ContainerNode cn, Subject caller, int depth) {
        // TODO: scan up to first output here??
    }

    public void write(OutputStream ostream) {
        this.output = new PrintWriter(new OutputStreamWriter(ostream));
        ...
    }
}
```
I would probably move as much state as possible inside this nested class, and maybe even the accumulate and output methods from the parent (eg the Writer declared above).

I think we could make this do a lazy init with a method in the nested class that scans/accumulates up to the point where it has to do output and then stops. At that point the HttpUpload is created and `run()`. I think doing that in the ctor would make usage very straightforward... this is typical of a dynamic Iterator (they advance to the first item in the ctor).

pdowler (changes requested):

cadc-vos-server/src/main/java/org/opencadc/vospace/server/async/RecursiveNodeSizeRunner.java
```diff
@@ -0,0 +1,335 @@
...
+        if (!authorizer.hasSingleNodeReadPermission(root, subject)) {
+            sendError("no read permission on the container node: " + nodePath);
+            return false;
+        }
+
+        // Write the report to a local temp file first, then upload it.
+        tmpFile = File.createTempFile("node-size-report-", ".txt");
+
+        try {
+            reportWriter = new BufferedWriter(new OutputStreamWriter(new FileOutputStream(tmpFile), StandardCharsets.UTF_8));
```
pdowler: when assigning to a class  member variable, I like to explicit use `this`. eg:
```
this.reportWriter = 
```
This is usually required in constructors and set methods anyway and in places like this it makes it clear where the declaration is and that class state is being directly changed. I generally do not explicitly use `this.` except in this case (left hand side of =).

In a syntax-highlighting IDE it doesn't add much, but it's nice in code reviews and when looking at source in other ways.

cadc-vos-server/src/main/java/org/opencadc/vospace/server/async/RecursiveNodeSizeRunner.java
```diff
@@ -0,0 +1,335 @@
...
+        request.version = VOS.VOSPACE_21;
+        request.getProtocols().add(sc);
+        List<Protocol> endpoints = nodePersistence.getTransferGenerator().getEndpoints(dest, request, null);
+        if (endpoints == null || endpoints.isEmpty()) {
+            throw new IllegalStateException("endpoint not found for: " + dest);
+        }
+        return new URL(endpoints.get(0).getEndpoint());
+    }
+
+    private void ensureDestDataNode(Subject caller) throws Exception {
```
pdowler: hmmm. Normally transfer negotiation creates the DataNode if it does not exist, but I guess this runner is bypassing that code. Yeah, looks like you pulled some code from PushToVOSpaceNegotiation... but not the call to `checkQuota`. That's interesting because this runner is intended to help people near or over their quota. OK for now but we should remember discuss.

anyway, this does represent some code duplication and it would be bad for these to diverge. Can you see if it is feasible to make `ensureDestDataNode` a method (public static?) in PushToVOSpaceNegotiation and re-usable here (eg extract lines 115-146 of PushToVOSpaceNegotiation into the method).

pdowler (changes requested):
This is very close. I think you need to pull some changes from main (recent cavern and recent cadc-test-vos changes). I am going to build and run so I can see what the output looks like.

